PR #32, opened 2016-08-24
Two origin-related additions

ianbjacobs:
1) Resolution of issue 27 is that payment app should have payee origin
information.
2) Add some security considerations about use of origin for
establishing app authenticity.


adrianhopebailie:
LGTM

PR #39, opened 2016-09-08
Numerous changes based on Rouslan review and issue migration

ianbjacobs:


index.html
```diff
@@ -301,33 +301,33 @@
     <ul>
       <li>Registration provides a way for browsers
 	to remain aware of the user's payment apps across transactions.</li>
-      <li>This design should not require registration as a
+      <li>Registration is not a 
 	prerequisite for using a payment app. In particular, a user
-	should be able to pay with a merchant-recommended payment app
+	should be able to pay with a payee-recommended payment app
```
adrianhopebailie, line 306: Is this true? I think what we want is the ability for an app to be registered on the fly so that a payee can recommend an app that isn't registered and the user can select it but it will still be registered before it is invoked.

ianbjacobs: Hi Adrian,
We discussed this on one of the payment apps calls. Shopify made the point that there may be cases of one-off usage of a payment app, in which case registration should not be mandatory. We concluded that registration is needed for persistence, but unnecessary for one-off usage.

Ian

adrianhopebailie: Okay cool, just wasn't sure if that was what was intended.

It would seem that means payment apps need to design for this and be able to process requests without assuming they are registered.

We should also consider the security implications of allowing this because it's nice to have heavy robust security during registration so that payment can be light.

ianbjacobs: Do you want to add any notes on these topics to the text?

adrianhopebailie: No, I wasn't on the call so can't complain now. Will raise an issue

ianbjacobs: I didn't mean to suggest that you were complaining or holding up the edit. I only meant that since you made some good comments, I didn't want to lose them, so should we also (e.g., after merging this edit) add them?

Ian

adrianhopebailie: > I didn't want to lose them, so should we also (e.g., after merging this edit) add them?

Good idea

adrianhopebailie:
LGTM

PR #40, opened 2016-09-12
Implications from non registration (via AHB)

ianbjacobs:
- ability to process without assuming prior registration
- security consideration


adrianhopebailie:
LGTM, thanks for adding

PR #52, opened 2016-09-26
Cleaned up text

halindrome:
The original wording had an extra "the" and was a little awkward


ianbjacobs:
I will take this into account in other edits.

PR #62, opened 2016-09-28
Editorial tweaks after Rouslan edits

ianbjacobs:
- removed payment options (which Rouslan had started). (However, left a
  note in about payment options as a hook for further discussion if
  needed). Also, left the _code_ about options in place; that is likely
  to need some harmonization as well.
- added back more clearly browser-based recommended apps
- Some editorial tweaks for language


ianbjacobs:
Fixed text based on task force discussion; left options in.

PR #68, opened 2016-11-06
Display order edits

ianbjacobs:
Based on 2 Nov 2016 discussion [1]

* distinguish bullets for matching payment apps from bullets for
recommended payment apps
* Define “ignored payment app” to include user configuration and
security issues
* Add note to first bullet of matching payment apps to remind people
that “matching” excludes “ignored”, which gives the user agent some
room to not display some matching payment apps
* Make clearer and strong (MUST) that user-side prefs override
payee-side prefs for matching payment apps.
* Also make clear that user-side prefs can be based on both user
configuration and usage patterns
* Editorial smoothing throughout, including in list of examples of
display behavior.

[1] https://www.w3.org/2016/11/02-apps-minutes.html


adrianhopebailie (approved):
LGTM

PR #70, opened 2016-11-16
Edited specification to remove notions of open/proprietary

ianbjacobs:
Issue 67 was about lack of definitions of proprietary and open;
https://github.com/w3c/webpayments-payment-apps-api/issues/67

Those terms were not necessary for the specification, so it was edited
to remove the concepts.

### DIFF
--- a/index.html
+++ b/index.html
@@ -753,8 +753,8 @@
           </li>
           <li>
             Register the <a>payment app</a> with the user agent for future use,
-            associating <var>manifest</var>'s <code>label</code> and
-            <code>icon</code> set with the payment app for user reference.
+            associating <var>manifest</var>'s <code>name</code> and
+            <code>icons</code> set with the payment app for user reference.
           </li>
           <li>
             For each <a>PaymentAppOption</a> present in the
@@ -763,7 +763,7 @@
               <li>
                 Add a new payment option to the <a>payment app</a>'s
                 registration, associating it with the <a>PaymentAppOption</a>
-                <code>label</code> and <code>icon</code> fields.
+                <code>name</code> and <code>icons</code> fields.
               </li>
               <li>
                 For each <a>payment method</a> indicated in the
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -324,9 +324,9 @@
...
-<li>    For privacy, the design should limit information about the user's environment available to merchant without user consent. That includes which payment apps the user has registered. For open payment methods, the merchant should not receive information about which payment app the user selected to pay unless the user consents to share that information. See <a href="https://github.com/w3c/browser-payment-api/issues/224">issue 224</a> for discussion about how merchant may track progress of a push payment.</li>
-<li>    Although decoupling relieves merchants of implementing some aspects of the checkout experience, one consequence is that they give up some degree of control. This was already the case for proprietary payment methods, but for open payment methods such as cards, merchants (or their PSPs) will be entrusting some portion of data collection to third party payment apps.</li>
+<li> User choice of payment apps that support a given payment method will depend on the payment method. For example, there may only be one app authorized to support a payment method owned by a company, while there may be many apps that support basic credit card payments or credit transfers.</li>
```
adrianhopebailie: Proposed:

> User choice of payment apps that support a given payment method will depend on the method and, specifically, if that method has a manifest that limits the apps that are allowed to process payment requests for that method.

index.html
```diff
@@ -324,9 +324,9 @@
...
+<li>    For privacy, the design should limit information about the user's environment available to merchant without user consent. That includes which payment apps the user has registered. The merchant should not receive information about which payment app the user selected to pay unless the user consents to share that information; of course when there is only one payment app for a given payment method that information is already publicly known. See <a href="https://github.com/w3c/browser-payment-api/issues/224">issue 224</a> for discussion about how merchant may track progress of a push payment.</li>
+<li>    Although decoupling relieves merchants of implementing some aspects of the checkout experience, one consequence is that they give up some degree of control. This was already the case for some payment methods, but for traditional card payments, merchants (or their PSPs) will be entrusting some portion of data collection to third party payment apps.</li>
```
adrianhopebailie: Propose:

> some portion of data collection to the browser or third party payment apps.

index.html
```diff
@@ -397,7 +397,7 @@
 <li>    A PAI should include origin information. This origin information may be used in a variety of ways:
   <ul>
 <li>        The origin information could enable user agents to provide the user with useful services when the user is browsing a site with the same origin (e.g., putting that payment app at the top of a list or otherwise highlighted).</li>
-<li>        For a "proprietary" payment method with an associated origin, the user agent can do some (security) checks by comparing the origin of the payment method and the payment app.</li>
+<li>        For a payment method with an associated origin, the user agent can do some (security) checks by comparing the origin of the payment method and the payment app.</li>
```
adrianhopebailie: Propose:

> For a payment method with an associated origin, and in the absence of a manifest, the user agent can do some (security) checks by comparing the origin of the payment method and the payment app.

index.html
```diff
@@ -1048,7 +1048,7 @@
 
     <p>For the display of recommended payment apps:</p>
     <ul>
-      <li>The user agent <span class='rfc2119'>SHOULD</span> display recommended payment apps.</li>
+      <li>The user agent <span class='rfc2119'>SHOULD</span> display recommended payment apps and configuration to not display recommended payment apps.</li>
```
adrianhopebailie: This sentence doesn't really make sense.

Propose:

> display recommended payment apps and **allow user** configuration to not display recommended payment apps

adrianhopebailie (changes requested):
Some suggested changes.

ianbjacobs:
@adrianhopebailie,

I liked your suggestions and pushed some updates as follows:
https://github.com/w3c/webpayments-payment-apps-api/pull/70/commits/50d5bb25534b78b6a69544b49797ab85d2133461

I don't think we should assume a manifest is the only mechanism here, so I've tweaked your suggestions to speak about payment method owner authorization (which, for example, might
be represented via a manifest file).

Ian

adrianhopebailie:
> I don't think we should assume a manifest is the only mechanism here, so I've tweaked your suggestions to speak about payment method owner authorization (which, for example, might
> be represented via a manifest file).

I think we should be explicit so that anything other than relationships between method and app managed by a manifest are outside our scope...

E.g.: Assuming there is some other mechanism for payment method X's owners to manage the relationship between their X and payment apps (like a special agreement with browser vendors that Y app is the preferred app for their payment method and Y and Z must never appear together).

If the behaviour that the browsers implement doesn't match what we have specified then they are not conforming to the spec. But, if the spec only defines behaviour for when a manifest is the source of the method/app relationship then they can implement these custom behaviours without being non-conforming.

adrianhopebailie (approved):

Issue #63, opened 2016-10-14
Should payment apps implement payment method-specific filters?


ianbjacobs:
Matt Saxon suggested to Ian that in the real world, payment method filtering is handled by payment apps, citing Apple Pay, Samsung Pay, Android Pay as examples.

One idea is to add to the payment app specification the following:
- Payment methods may define filtering capabilities.
- Merchants would include filter information in the 'data' object (of payment request API) for accepted payment methods
- Payments apps would be responsible for implementing matching using the filter information. Payment apps could choose not to implement filtering for a given payment method.
- There would be a standard function implemented by payment apps that would respond true or false to queries from the mediator about (filter) matching for each payment method.

Thus, payment method-specific filters would be implemented by payment apps, not the mediator.

Note: For in-browser basic card support, this can also be modeled as the browser-as-payment-app implementing matching rather than the browser-as-mediator.

adamroach:
To capture some notes about what we discussed on the payment apps call -- there were concerns (both performance-related and privacy-related) about the proposal as stated. One proposal was that part of app registration could include a lambda function -- required to be a _pure function_ -- that took the filter as input and generated a boolean "I can handle this" as output. These would be evaluated in an isolated context with a null origin (or what specs mean when they say the "origin is a globally unique identifier"), which would prevent external access (addressing the performance concerns) and prevent local storage (which would, in concert with preventing external access, address the privacy concerns).

ianbjacobs:
Adding notes from 19 October app task force discussion:
https://www.w3.org/2016/10/19-apps-minutes.html#filtering
- Need to clarify whether the expectation is that the mediator will still match on the payment method identifier and the payment app will only handle the filter.
- AdamR pointed out that up to now, payment apps have only known about a payment request once selected by the user. But if they are polled for filter matching, this could (depending on how it's proposed), leak information to the app about the existence of a payment request BEFORE the app has been selected.
- Jason points out that some payment apps might implement filtering on the server side, and performance would be a concern.
- AdamR began the thought process of whether a function that returns a boolean about filter matching could be a pure function provided at registration, and that could be run without letting the payment app know.

ianbjacobs:
At the 2 Nov call, Adam Roach took an action to work on the notion of an isolated function:
https://www.w3.org/2016/11/02-apps-minutes.html#item03

rsolomakhin:
> Generate a boolean "I can handle this" as output.

[...]

PR #76, opened 2016-12-14
Lambda functions for app matching

adamroach:
This is still pretty rough, and not intended to be merged as-is. It is intended to be useful as the basis for discussing whether this is the right approach to handle Issue #63. It will need further refining if we decide to pursue this approach.

### DIFF
--- a/index.html
+++ b/index.html
@@ -834,7 +834,12 @@
   </h2>
   <p class="issue"><a href="https://github.com/w3c/webpayments-payment-apps-api/issues/69">Issue 69</a>: The Payment Apps Task Force has a goal of alignment with the draft <a href="https://github.com/w3c/manifest">Web App Manifest</a> specification for data used to display payment app and options for selection by the user. More work is necessary to determine whether the Payment App API should reference (part of) the Web App Manifest specification, import definitions while that specification remains a draft, or define new terms.</p>
       <pre id="payment-app-manifest-idl" class="idl">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1011,6 +1041,106 @@
...
+          not considered to be a candidate to handle the
+          <code>PaymentRequest</code> and terminate these steps.
+        </li>
+        <li>
+          If the <code>PaymentAppManifest</code> does not contain a
+          <code>canHandle</code> attribute, then this <a>payment app</a> is a
+          candidate for handling the <code>PaymentRequest</code>. Only those
+          <a>PaymentAppOption</a>s that have <code>enabledMethods</code>
+          containing at least one member of <var>commonMethods</var> are to be
+          displayed to the user. Terminate these steps.
```
rsolomakhin, line 1097: Can we rephrase "to be displayed to the user" as "to be made available to the user for selection"? Chrome displays only one PaymentAppOption at first, with others available in a dropdown.

rsolomakhin (commented):
This looks good, sans our conversation about returning a sequence of dictionaries instead of an object.

ianbjacobs:
+1 to "to be made available to the user for selection"

adrianhopebailie:
Really like this feature! Thanks @adamroach and @rsolomakhin for putting it together

dlongley:
+1

PR #80, opened 2017-01-05
Edits based on recent discussions and decisions

ianbjacobs:
This was not a comprehensive review of the document. However, I
reviewed it in light of some recent decisions and did some updates:

* Given 4 January discussion about recommended payment apps, I reviewed
the text on that topic and made changes.
* I also cleaned up some text on registration given the new assumption
that registration IS a prerequisite for usage of a payment app that
conforms to this specification.
* It has also become clear that this specification is not designed to
address native payment apps, and so I deleted some text on native
payment apps, and also stated more clearly that those mechanisms lie
outside the scope of this document.
* IMPORTANT: It is my current understanding that a payment app
identifier will designate service worker code. Therefore, the spec now
says that; we’ll discuss whether this was the right edit at our 10
January call.
* I tried to reduce instances of the word “display” in light of recent
discussions about “enabling the user” rather than always talking bout
display of information. Nonetheless, some instances of “display” still
remain in the spec where they make sense.
* I added mention of HTTP Link headers for finding payment app manifest
files; we may or may not need that but I’m keeping this spec aligned
with the payment method stuff.
* I added mention of paymentRequestID and otherwise cleaned up
discussion of reconciliation.
* Now that we have canHandle I removed some other notes and provided a forward reference to it.
* We also need to figure out (cf issue 79) how and when the payee provides information about recommended payment apps.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1406,24 +1406,25 @@
   <h3>
     Payment App Display
   </h3>
-    <p>
-      Payment Apps that require user input can open a payment window using the
-      <code>clients.openWindow()</code> method defined in [[!SERVICE-WORKERS]].
-      Absent user preferences that override such behavior, user interaction is
-      required during payment requests, in the form of payment app selection. As
-      a consequence, the user agent MUST treat a <a>paymentrequest</a> event as
-      user interaction for the purposes of determining whether the <a>service
-      worker</a> is allowed to open a window.
-    </p>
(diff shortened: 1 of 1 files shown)


ianbjacobs:
I am merging this PR after brief email with AHB. There are differing views on the PAI, so I've added a pointer to issue 48 and we'll use the revised text as we continue the discussion.

PR #84, opened 2017-01-12
Introduce the PaymentAppResponseData dictionary

tommythorsen:
This is used instead of PaymentResponse to pass data back from the
payment app to the user agent.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1425,6 +1425,28 @@
       </ol>
     </section>
   </section>
+  <section id="sec-app-response">
+    <h3>Payment App Response</h3>
+    The <a>payment app response</a> is conveyed using the following
+    dictionary:
+    <pre class="idl">
+      dictionary PaymentAppResponse {
+        DOMString methodName;
+        object details;
+      };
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1426,6 +1426,28 @@
       </ol>
     </section>
   </section>
+  <section id="sec-app-response-data">
+    <h3>Payment App Response Data</h3>
+    The <a>payment app response data</a> is conveyed using the following
+    dictionary:
+    <pre class="idl">
+      dictionary PaymentAppResponseData {
```
dlongley: Should the name be simply `PaymentAppResponse` instead? Does appending `Data` to it add anything?

tommythorsen: I suppose that would be just as good a name, but then I would insist that we also change `PaymentAppRequestData` to `PaymentAppRequest`

ianbjacobs: I don't feel strongly about "Data." (I do want to avoid confusion with "Payment Request" and "Payment Response")

Issue #96, opened 2017-01-23
Revisiting payment app filtering


adamroach:
The current spec includes a function that a payment app can optionally register for itself (`canHandle()` in the existing `setManifest()` method) for the purpose of determining whether the payment app is applicable to the payment method _and its filters_. This is one of several mechanisms that have been discussed to date.

@marcoscaceres correctly points out that, as specified, the current mechanism is somewhat problematic, in that we're assuming basically a "pickling" of this function for use in future contexts in a way that's not obviously supported by ES.

In practice, when I proposed this approach (and, to be clear, it was my second-favorite choice out of options on the table), what I had in mind was something very similar to how [PAC files](https://en.wikipedia.org/wiki/Proxy_auto-config) operate, both from a persistence and execution environment perspective. While there are probably ways to specify `canHandle()` in a way that is more palatable (e.g., by having it point to an external file that is evaluated independently), there is some question as to whether doing so is worth the effort it is likely to take.

The alternate approach is to more concretely define a payment filtering algorithm based on a generic understanding of the `filter` parameter passed as part of the `PaymentRequest`.

I'm opening this issue to re-kindle the discussion regarding these two potential approaches, in light of the newly highlighted difficulties we are likely to encounter. Of course, alternatives approaches beyond the two I describe are welcome as well.

jakearchibald:
If this cannot be determined from data the browser already has, and needs to execute code in the origin, a service worker event is the best fit.

tommythorsen:
I think that changing this into an event might not be such a good idea privacy-wise. In a regular event handler, there's nothing stopping the service worker from logging all of the payment request information and passing it back to the back-end. That would let an information-greedy payment provider see all of the user's potential payments -- even the ones that don't ultimately end up in that particular payment app.

The intention of the currently specified mechanism is that the registered `canHandle()` function should be a "pickled", "pure" function ("pure" meaning that the function does not rely on anything other than the input parameters, and does not have any side-effects).

I'm not actually a great fan of the current `canHandle()` mechanism, but since I don't have a better suggestion, I think it's acceptable, as long as we can show that it's implementable. If not, then I would like to propose that we drop the whole `canHandle()` mechanism.

[...]

Issue #95, opened 2017-01-23
Replace setManifest()/getManifest() with set()/get()/keys()/has()/delete()


adamroach:
When a payment app is being set up in the current spec, it does so by (a) registering as a service worker, and then (b) installing a monolithic object via `PaymentAppManager.setManifest()`. Subsequent updates are performed through calling `PaymentAppManager.getManifest()`, mutating the object, and setting it again.

@marcoscaceres has proposed that this be replaced with a decomposed interface that allows for payment apps to modify their characteristics on a more granular level using get/set/delete/etc. operations.

As a slight modification of the proposal he made in his payment-request-handler document, I propose concretely:

1. Removing the `setManifest()` and `setManifest()` methods from `PaymentAppManager`.
2. Adding new `setName()` and `setIcon()` methods to `PaymentAppManager`.
3. Adding a new `options` attribute to `PaymentAppManager`, of type `PaymentMethodOption`, having the following definition (taken from Marcos' proposal):

```webidl
[Exposed=(Window,Worker)]
interface PaymentMethodOptions {
    [NewObject]
    Promise<boolean>              delete(DOMString methodKey);
    [NewObject]
    Promise<PaymentMethodDetails> get(DOMString methodKey);
    [NewObject]
    Promise<sequence<DOMString>>  keys();
    [NewObject]
    Promise<boolean>              has(DOMString methodKey);
    [NewObject]
    Promise<void>                 set(DOMString methodKey,
                                                      PaymentMethodDetails details);
};
```

tommythorsen:
I think this looks like a good change. +1 from me

marcoscaceres:
> Adding new setName() and setIcon() methods to PaymentAppManager.

I still don't understand the use case for these? Can you clarify?

jakearchibald:
@marcoscaceres for the payment app to be represented to the user, I believe. You have the name & icons per method, the current spec has one name & one set of icons for the app.

marcoscaceres:
Isn't that just duplicating HTML (link rel=icon and meta application-name) and web manifest? Why do we need YA another way of doing this? (That's a question for the group, not Jake).

marcoscaceres:
Also, we need multiple icons for per payment method: those actually matter, but should still be optional. 

The app should get it's icon, name, etc. like any other PWA.

ianbjacobs:
@marcoscaceres wrote:

[...]

PR #104, opened 2017-02-20
Proposed changes for #95 (and #96)

adamroach:


index.html
```diff
@@ -611,10 +618,153 @@ <h3 id="summary-recommended-apps">Recommended Payment Apps</h3>
...
+        selection.
+      </dd>
+    </dl>
+  </section>
+  <section id="payment-options">
+  <h2>
+    <a>PaymentOptions</a> interface
+  </h2>
+      <pre id="payment-app-manager-idl" class="idl">
+      interface PaymentOptions {
```
bentian: The interface name conflicts with [PaymentOptions dictionary](https://w3c.github.io/browser-payment-api/#paymentoptions-dictionary). May this get confusing?

index.html
```diff
@@ -960,7 +1090,7 @@ <h3 id="summary-recommended-apps">Recommended Payment Apps</h3>
         </li>
         <li>
           For each <a>PaymentAppOption</a> <var>option</var> in the
-          <a>payment app</a>'s <a>PaymentAppManifest</a>, add all entries in
+          <a>payment app</a>'s <a>PaymentAppAManager.options</a>, add all entries in
```
bentian: an extra 'A' here

tommythorsen:
Replacing `setManifest`/`getManifest` with the `options` attribute seems like a good change, but I'm not so sure about the `wallets` attribute.

While we have talked about the usefulness of being able to let a single origin contain multiple "Apps" that again can contain multiple "Options", I don't necessarily think that containing this entire hierarchy inside of a single Service Worker is a good idea. Will each of these "Wallets" typically be presented to the user as an "App"? Or will they again be grouped inside another "App" container?

adamroach:
@tommythorsen -- The notion here is to allow payment app developers to structure their service workers in whatever way makes sense for them, independently of how they wish options to be presented to the users. The current proposal would allow multiple service workers to handle different options within the same wallet, or a single service worker to handle multiple wallets. There seems no reason to artificially constrain the way applications structure their service workers just to change how things are presented to users.

In terms of rendering, what I would *expect* -- since wallets are optional -- is that browsers would show app-level information for payment apps without wallets, and wallet-level information for payment apps with. They would appear "at the same level," so to speak.

So, for example, if American Express had registered two wallets (one app!), and Bank of America had registered one app (but not used the wallet concept at all), it would be reasonable to render this to the user as:
wallets

tommythorsen:
@adamroach: In the screenshot above, have you given any thought to where the "Bank of America" string and the red Bank of America icon should come from? I'm interested in starting to implement some of these things, or at least write some example code, but I have a hard time nailing down where "App" level information should exist, and even [what a "Payment App" really is](https://github.com/w3c/webpayments-payment-apps-api/issues/105).

Maybe we should specify "App" level information from the Service Worker too? Maybe something like:

```javascript
interface PaymentAppManager {
    attribute PaymentAppDetails app;
    attribute PaymentAppOptions options;
    attribute PaymentWallets wallets;
};

dictionary PaymentAppDetails {
    DOMString appKey;
    required DOMString name;
    sequence<ImageObject> icons;
};
```

adamroach:
@tommythorsen --
> In the screenshot above, have you given any thought to where the "Bank of America" string and the red Bank of America icon should come from?

The notion here is that the service worker has been registered from some context (i.e., an HTML page), and that context provides information specific to the web app. I'm hardly an expert here, but my understanding is that modern web browsers get the name from:

- The Web App Manifest indicated in the HTML via `<link rel="manifest" href="`*name of manifest file*`">`
- If that is not present, indicated in the HTML via `<meta name="application-name" content="`*app name*`">`
- If neither is present, via `<title>`*app name*`</title>`

Similarly, the icon comes from:
- The Web App Manifest indicated in the HTML via `<link rel="manifest" href="`*name of manifest file*`">`
- If that is not present, indicated in the HTML via `<link rel="icon" href="`*icon location*`">`
- If neither is present, from `/favicon.ico`

To be clear, I'm not proposing this *per se*; it is my understanding that this is how Web Apps in general are assigned names and icons for presenting to users. If we decide that this model isn't appropriate for payments, I don't *personally* have a problem with changing it -- but if we have an already-established pattern and set of tools for this kind of thing, we'll need some clear reasons for doing things another way.

tommythorsen:
@adamroach I agree with that way of doing it. I think that the only time this information is available to us, is at the time when we register the Service Worker. We'll probably need to pick the information out of the browsing context somehow, and store it along with the other information in the PaymentAppManager.

What do we do if the Payment Provider would like to rebrand, and change the name and icon that belongs to a Payment App? Aren't we going to need some way to update these from the installed service worker?

ianbjacobs:
@tommythorsen wrote:

"What do we do if the Payment Provider would like to rebrand, and change the name and icon that belongs to a Payment App? Aren't we going to need some way to update these from the installed service worker?"

Wouldn't that be taken into account the next time I visit the site? (There may be a separate issue about how caching is managed for this information, but that seems out of scope for our spec.)

Ian

adrianhopebailie:
@tommythorsen wrote:

"What do we do if the Payment Provider would like to rebrand, and change the name and icon that belongs to a Payment App? Aren't we going to need some way to update these from the installed service worker?"

Isn't this up to browsers to define? Do we need to specify anything as relates to UI? The Service Worker life cycle is already well-defined.

tommythorsen:
@ianbjacobs 
> Wouldn't that be taken into account the next time I visit the site? (There may be a separate issue about how caching is managed for this information, but that seems out of scope for our spec.)

Well, each page within the site can have different icons and names. It would be feasible to update this information every time the user visits the exact same web page from which they originally registered the service worker. *However*, how likely is it that they will ever visit that exact page again? The service worker was probably registered from some kind of a "signup" page. Why would the user go there again?

@adrianhopebailie 
> Isn't this up to browsers to define? Do we need to specify anything as relates to UI? The Service Worker life cycle is already well-defined.

I think this is up to us to define. We have introduced these names and icons, and it's up to us to define their entire life cycle.

It would be nice if we could inherit this behavior from the Web App Manifest specification, but [that particular section](https://w3c.github.io/manifest/#updating) is not very helpful. In light of this, it doesn't seem unreasonable to postpone thinking about updating this information until after v1 of our specification, though.

dlongley:
> I think this is up to us to define. We have introduced these names and icons, and it's up to us to define their entire life cycle.

Can we spec it such that when the service worker is updated, the icons/other information will be updated as well?

adrianhopebailie:
>> I think this is up to us to define. We have introduced these names and icons, and it's up to us to define their entire life cycle.
> 
> Can we spec it such that when the service worker is updated, the icons/other information will be updated as well?

My point is, I don't think we can even define how the browser gets the icons and labels without treading in dangerous security territory.

As a develop I can register a bunch of options that I want the user to be prompted with but ultimately, how those are rendered to the user is up to the browser. Will they have icons and labels at all?

Before we spend too many cycles figuring out how to specify the icons and labels let's agree that the browser will use them.

As a sidebar, and illustration of the security risks, PayPal is being phished extensively by scammers that have got SSL certs from LetsEncrypt and thereby showing the "green bar of safety" to users.

Any visual affirmation a user gets that a payment app published by a reputable brand should be carefully controlled, especially if it is shown inside browser generated UI.

PR #108, opened 2017-03-15
Run the spec through HTML5-Tidy.

tommythorsen:
This re-indents and re-wraps the specification. I have also fixed some warnings that were output by the tidy tool.


ianbjacobs (approved):
(I am not reviewing individual tidy changes but am ok with running tidy)

PR #110, opened 2017-03-17
Per 16 March teleconf [1], displayItems are not handed to the payment app

ianbjacobs:
app.

[1] https://www.w3.org/2017/03/16-wpwg-minutes.html#item04


adamroach (approved):
Looks good to me.

PR #114, opened 2017-03-20
Small fixes

ianbjacobs:
* Small fixes
* Also clarified that paymentResponse is not necessarily for successful
completion of "payment" in the sense of exchange of funds. For example,
it might involve merely returning card credentials. So reframed in terms
of successful user interaction.

Question: can/should we reuse PaymentResponse rather than create a new PaymentAppResponse?

### DIFF
--- a/index.html
+++ b/index.html
@@ -758,7 +758,7 @@
     Request API, the user agent displays a list of matching origins
     for the user to make a selection. This specification includes a
     limited number of display requirements; most user experience
-    details are left to user agent implementers.</p>
+    details are left to implementers.</p>
     <section>
       <h3>Ordering of Payment Handlers</h3>
       <ul>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -758,7 +758,7 @@
     Request API, the user agent displays a list of matching origins
     for the user to make a selection. This specification includes a
     limited number of display requirements; most user experience
-    details are left to user agents.</p>
+    details are left to user agent implementers.</p>
```
marcoscaceres: maybe just "left to implementers"

index.html
```diff
@@ -1189,8 +1189,8 @@
           Promise.</p>
           <p>If the Promise is rejected, the user agent MUST run
           the <dfn>payment app failure algorithm</dfn>. The exact
-          details of this algorithm is left for user agent
-          implementers to decide on. Acceptable behaviors include,
+          details of this algorithm are left to user agent
+          implementers. Acceptable behaviors include,
```
marcoscaceres: Same as above.

marcoscaceres (changes requested):

marcoscaceres (approved):

PR #121, opened 2017-03-28
The instrumnets/wallets attributes should be readonly and [SameObject].

romandev:
The attributes should be readonly because the PaymentInstruments and
PaymentWallets doesn't have constructor and developers can not set.
Also, it's better to mark [SameObject] because we don't need to create a new
object whenever accessing those attributes.

### DIFF
--- a/index.html
+++ b/index.html
@@ -338,8 +338,8 @@
       <h2><a>PaymentManager</a> interface</h2>
       <pre id="payment-manager-idl" class="idl">
       interface PaymentManager {
-        attribute PaymentInstruments instruments;
-        attribute PaymentWallets wallets;
+        [SameObject] readonly attribute PaymentInstruments instruments;
+        [SameObject] readonly attribute PaymentWallets wallets;
       };
       </pre>
       <dl>


romandev:
@ianbjacobs Could you please review this? (I found this bug when implementing this feature in user agent side.)

ianbjacobs:
Hi @ romandev, 

Thank you for logging this issue. I would be more confident if @adamroach were to review the proposal. 

Ian

adamroach:
@ianbjacobs  -- The assertions sound sensible, but we're getting into deep WebIDL guts here, somewhat beyond my expertise. @marcoscaceres? Care to comment?

marcoscaceres (approved):
Seems fine. These are singleton references and don't change for the life of the application.

ianbjacobs:
@marcoscaceres, I assume that even though these "don't change for the life of the application" the app distributor can still add/remove wallets and instruments.

marcoscaceres:
Correct.

ianbjacobs:
Marked as non-substantive for IPR from ash-nazg.

PR #130, opened 2017-04-13
Add support for topLevelOrigin and paymentRequestOrigin

ianbjacobs:
Per task force discussion about origin:
 https://www.w3.org/2017/04/04-apps-minutes#item01

Also updated open window algorithm, which referred to the wrong origin.
Now checks that window origin same as payment app service worker client origin.

### DIFF
--- a/index.html
+++ b/index.html
@@ -415,7 +415,7 @@
             <ol>
               <li>Resolve <var>p</var> with a <dfn>Sequence</dfn> that contains all
               the <var>instrumentKey</var>s for the <a>PaymentInstrument</a>s
-              contained in the collection.
+              contained in the collection, in original insertion order.
             </ol>
             <li> Return <var>p</var></li>
           </ol>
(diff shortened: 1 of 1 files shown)


frank-hoffmann:
Should we specify the expected value of `paymentRequestOrigin` if the PaymentRequest is not initiated through an iframe or is it clear that both `topLevelOrigin` and `paymentRequestOrigin` will have the same value in these cases?

ianbjacobs:
The intent is that both will always have values, and the values will differ in the iframe case. I would be fine to add a sentence to that effect, so paymentRequestOrigin would become:

This attribute is a string that indicates the <a data-cite="!HTML5#origin">origin</a> where a <a>PaymentRequest</a> was initialized. When a <a>PaymentRequest</a> is initialized in the <a>topLevelOrigin</a>, the attributes will have the same value, otherwise the attributes will have different values. For example, when a <a>PaymentRequest</a> is initialized within an iframe from an origin other than <a>topLevelOrigin</a>, the value of this attribute is the origin of the iframe. 
The string MUST be formatted according to the "<a data-cite="!RFC6454#section-6.1">Unicode Serialization of an Origin</a>" algorithm defined in section 6.1 of [[!RFC6454]].

rsolomakhin (approved):
LGTM % adding a note about `paymentRequestOrigin` value when not invoked inside of an iframe.

rsolomakhin (approved):
LGTM

PR #138, opened 2017-04-17
Editorial changes for FPWD

adrianhopebailie:


index.html
```diff
@@ -362,13 +361,12 @@
...
         "PaymentManager.wallets">wallets</a>.
         </li>
-        <li>Payment apps make use of <a>service worker</a>s, starting with
-        registration of their capabilities with the user agent. A service
-        worker registration has zero or one <a>payment handler</a>. However, an
-        origin may implement a payment app with more than one service worker
-        registration.
+        <li>Payment apps make use of <a>service worker</a>s which must define a
+        <a>payment handler</a> to handle the <a>PaymentRequestEvent</a>s that
+        are raised by the user agent.
```
ianbjacobs: I have several questions about your proposed change:

 * As we are still in the introduction, I would like to keep the fact that it is through registration that a user agent learns about (some) payment app capabilities: those related to payment handling.
* It is not clear to me that user agents "raise" these events. Rather, they result from the PR API invocation.
* It is not clear from your edit that there can be more than one payment handler

How about instead:

<li>Payment apps make use of <a>service worker</a>s to register their payment handling capabilities with the user agent. This is done through one or more <a>payment handler</a>s, which listen to <a>PaymentRequestEvent</a>s raised through the Payment Request API.</li>

index.html
```diff
@@ -362,13 +361,12 @@
...
         </li>
-        <li>Registration associates with each payment handler:
+        <li>During registration the service worker will use the
+        <a>PaymentManager</a> to set:
```
ianbjacobs: I propose we say "the service worker uses" instead of speaking in the future.

index.html
```diff
@@ -388,36 +386,48 @@
...
-        <li>The user agent displays matching payment handlers, displaying and
-        grouping Instruments according to information (labels and icons)
-        provided at registration or otherwise available from the Web app.
+        <li>The user agent displays a set of choices to the user. These choices
+        are <a data-lt="PaymentManager.instruments">instruments</a> that were
+        registered through the <a>PaymentManager</a> of candidate payment
+        handlers, displaying and grouping the choices according to information
+        (labels and icons) provided at registration or otherwise available from
+        the Web app.
         </li>
```
ianbjacobs: Proposed simplification:

<li>The user agent displays a set of choices to the user: the registered
<a data-lt="PaymentManager.instruments">instruments</a> of candidate payment
handlers. The user agent displays and groups these choices according to information
(labels and icons) provided at registration or otherwise available from the Web app.</li>

index.html
```diff
@@ -388,36 +386,48 @@
...
         <li>Once activated, the payment handler performs whatever steps are
-        necessary to authenticate the user, process the payment, and return
-        payment information to the <a>payee</a>. If interaction with the user
-        is necessary, the <a>payment handler</a> can open a window for that
+        necessary to handle the payment request, and return an appropriate
+        payment response to the <a>payee</a>. If interaction with the user is
+        necessary, the <a>payment handler</a> can open a window for that
         purpose.
         </li>
```
ianbjacobs: Please leave "authenticate the user" in this description. We are frequently asked whose job it is to manage authentication, and I would like for this text to help answer the question.

index.html
```diff
@@ -512,57 +520,57 @@
...
-            part of the payment request flow, payment handlers must have at
-            least one registered payment Instrument, and that Instrument needs
-            to match the payment methods and required capabilities specified by
-            the payment request.
+            associated with a service worker (and therefor its payment
+            handler). To be invoked by the user as part of the payment request
+            flow, payment handlers must have at least one registered payment
+            instrument to present to the user. That instrument needs to match
+            the payment methods and required capabilities specified by the
+            payment request.
```
ianbjacobs: Notes:
* s/therefor/therefore
* "To be invoked by the user as part of the payment request flow"
   I believe earlier you refer to them as candidate payment handlers. So would it make sense to say:
   "To be a candidate payment handler, a handler must have at least one registered payment instrument..."
* s/specified by the payment request/of the <a>PaymentRequest</a>

index.html
```diff
@@ -512,57 +520,57 @@
...
           </p>
         </section>
         <section>
           <h2>
             <dfn>wallets</dfn> attribute
           </h2>
           <p>
-            This attribute allows payment handlers to group payment instruments
+            This attribute allows service workers to group payment instruments
```
ianbjacobs: Actually, in light of your previous edit, do you want to say "allows payment managers" ?

index.html
```diff
@@ -512,57 +520,57 @@
...
             payment handlers use this grouping mechanism, then matching payment
             instruments that do not appear in any groups should still be
-            presented to users by the browser for selection.
+            presented to users by the user agent for selection.
           </p>
         </section>
         <section>
           <h2>
-            <dfn>requestPermission()</dfn> method
+            <dfn>requestPermission</dfn> method
```
ianbjacobs: I saw that too, thanks for including it!

index.html
```diff
@@ -512,57 +520,57 @@
...
           </h2>
-          <p class="issue" title="Issue 94 - Permission">
-            <a href=
-            "https://github.com/w3c/webpayments-payment-apps-api/issues/94">Issue
-            94</a>. The means for code requesting permission to handle payments
-            is not yet defined.
+          <p class="issue" title="Permission" data-number="94">
+            The means for code requesting permission to handle payments is not
+            yet defined.
```
ianbjacobs, line 574: I have a question: you removed the link. Does your code provide a link to the issue? If so, thanks for teaching me how to do that. If not, please restore the direct link to the issue.

index.html
```diff
@@ -512,57 +520,57 @@
...
-          <a href=
-          "https://github.com/w3c/webpayments-payment-apps-api/issues/98">Issue
-          98</a>: Should the API support providing grouping information
-          ("wallets") to the user agent? What should requirements be on user
-          agents to display or provide for user interaction with this
-          information?
+        "Support grouped display of payment instruments?" data-number="98">
+          Should the API support providing grouping information ("wallets") to
+          the user agent? What should requirements be on user agents to display
+          or provide for user interaction with this information?
```
ianbjacobs, line 585: Same comment about direct link (and I won't repeat this comment for any later instances).

index.html
adrianhopebailie: This is intentional. I think that as a developer it makes sense for this spec to start by telling me how I define a payment handler. After that it explains how I register the payment methods and capabilities I support in a way that will allow my payment handler to be invoked.

These are two separate things I must do as a developer so I don't want to try and bundle them together.

i.e. Start with the payment handling logic then the glue that links the payment request, the user choice and the event than my logic handles together.

I have added a note later on about multiplicity. I think the zero or one language is confusing. While it's technically possible to register a service worker and set a bunch of instruments and not register a payment handler that's a pointless exercise.

I'd be interested to hear from implementors as we begin fleshing out the algorithms whether it makes sense for the PaymentManager to be disabled if there is no event listener for the `onpaymentrequest` event.

index.html
adrianhopebailie: The original text is incorrect. We should not bring discussion of authentication of the user into this text at all. 

IMO this is what causes the confusion and results in long mail trails from people who want to understand how we are handling authentication. The answer is simple, we are not. That is the job of payment apps.

I will add a non-normative note that explains what "handle a payment request" means and provide examples using basic card and credit transfer to explain

index.html
adrianhopebailie: Hmmm... I don't consider the payment manager as an active component but more of a data store.  The service worker is the active component and it has a payment manager which is the component it uses to do this.

index.html
adrianhopebailie: > Does your code provide a link to the issue?

Yes! Respec understand GitHub issues. In fact if you put an issue marker with no text in the body it will fetch the issue description from Github too. There is an example somewhere in this doc.

index.html
adrianhopebailie: > t is not clear to me that user agents "raise" these events. Rather, they result from the PR API invocation.

They do. There are two things happening here.

  1. The user agent handles the PR API invocation
  2. The user agent raises a `PaymentRequestEvent`

Technically these can both happen without the other and in future that may be the case. The browser may have other APIs or processes that trigger a `PaymentRequestEvent` in future

index.html
ianbjacobs: Given that the attribute is on the PaymentManager interface, however, it made sense to me that it relates to payment managers.

ianbjacobs (approved):
Ook!

index.html
dlongley: To throw another wrench in here ... perhaps `PaymentManager` should become `PaymentInstrumentStore` or `PaymentInstruments` or similar if it's not "managing" anything?

Regarding the original point, I don't think "service workers" are the ones doing the grouping, developers are -- they're the ones with agency here. Perhaps avoid the problem all together with:

`This attribute allows related payment instruments to be logically grouped together.`

or

`This attribute is used to group related payment instruments.`

PR #145, opened 2017-05-02
Fix to prevent duplicate ids by respec

ianbjacobs:
If you reference an issue mulitple times in the spec, for example:

<p class="issue" data-number="98"></p>

respec generates duplicate ids. So this edit removes that construct
in favor of some explicit links.

### DIFF
--- a/index.html
+++ b/index.html
@@ -976,7 +976,7 @@
             <li>Return <var>p</var>
             </li>
           </ol>
-          <p class="issue" title="Clear()" data-number="129"></p>
+          <p class="issue" title="Clear()">See <a href="https://github.com/w3c/webpayments-payment-handler/issues/129">issue 129</a>.</p>
         </section>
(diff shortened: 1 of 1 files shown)


adrianhopebailie:
Wouldn't it be better if we just had each issue only once?

@marcoscaceres - are there other side effects to having the issue numbers in the spec? i.e. Is there anything keeping a list of these?

adrianhopebailie (approved):

marcoscaceres:
There shouldn't be other issues.. if you are hitting duplicate issue problems, then the problem is that the issue itself should probably be split into 2?

ianbjacobs:
It may be possible to split the issue into multiple pieces. However, I think it is also the case that the issue manifests itself in multiple places in the spec. For now I have left a single issue with several hooks in the spec.

Issue #157, opened 2017-05-12
Revisiting Payment Method specific filter/capability matching


ianbjacobs:
Hi all,

At yesterday's meeting [1] we returned to the question of whether matching of payment method specific filters/capabilities happens in the user agent or the payment app. There were two topics:

 - How capabilities are identified (in data from the merchant)
 - Where matching takes place.

This issue is (for now at least) about the latter.

I think there is consensus that Payment Request API should indicate in relevant algorithms  THAT additional matching can take place according to payment method. I anticipate that PR API will not say WHERE matching takes place (user agent or payment app).

The current Payment Handler spec supports registration of payment app capabilities with the user agent with, I believe, the assumption that the user agent will do the matching. There are pros to this approach, including:

 * Payment apps don't have to implement matching
 * Payment apps do not learn about users transactions until selected by the user

There are also cons to this approach:

 * Simple string matching may not be adequate to address some use cases
 * Payment apps may not want to share user data with user agents
 * There may be challenges to keeping registration data up-to-date, especially if there
    are multiple instances of the payment handler (on different devices) that need to be updated.

After FPWD, I would like the payment apps task force to revisit this question of matching
to see whether we still wish to put forward an approach where matching is implemented
by the user agent, or more of a canMakePayment() approach where the payment app
returns a boolean (and where we are sensitive to privacy issues).
   
Ian

P.S. I started a new issue but this is also a continuation of some parts of the
closed issue 96 
  https://github.com/w3c/payment-handler/issues/96

[1] https://www.w3.org/2017/05/11-wpwg-minutes.html#item02

rsolomakhin:
After discussing this with @zkoch, I'm starting to lean toward the `.canMakePayment()` approach for the payment handler to perform the filtering. The user should have ability to avoid this call either via user agent setting or by entering private browsing mode. This alleviates the concern about leaking sensitive information to the payment handler before selection.

Note that the user must install the payment handler first before the handler is able to receive the `.canMakePayment()` call. This indicates elevated trust on the part of user, which also alleviates the privacy concern somewhat.

ianbjacobs:
@rsolomakhin, thank you for the follow-up from yesterday's call. In terms of the spec, this suggests the following sorts of changes:

 * In the PaymentInstrument dictionary, replace "capabilities" with a canMakePayment() method 
    (and defining it).
 * Including either a requirement or a strongly worded suggestion that the user agent provide
[...]

Issue #117, opened 2017-03-24
Support for Abort() being delegated to Payment Handler


mattsaxon:
In PaymentRequest, the statement "A user agent might not always be able to abort a request. For example, if the user agent has delegated responsibility for the request to another app. In this situation, abort() will reject the returned Promise."

Can the abort be delegated to a payment handler interface such that this statement invalid?

rsolomakhin:
Let's make this only optional, please.

mattsaxon:
I think they would be fine, especially if this linked in with the state model such that when the status is 'processing' that calling abort() results in an exception. I could see additional states in his state model e.g. Interactive->pre-processing->processing which could more clearly specify the behaviour and set better expectations for users of the API.

mattsaxon:
Related to [Issue 473 in PR](https://github.com/w3c/browser-payment-api/issues/473)

adrianhopebailie:
Per discussion on the Payment Apps TF call:

* We can't provide any guarantees that, once a payment request is delegated beyond the browser, it can be aborted. Web-based payment apps are only one type of app.
* We can improve the Payment Handler specification to include a mechanism of passing the abort request to a web-based payment app.

## Proposal

Add a `paymentRequestAborted` event to the Payment Handler interface. The event will have a `respondWith` method that takes a boolean parameter indicating if the paymentRequest has been successfully aborted.

rsolomakhin:
I've head support for this from payment app developers. There should be no expectation of rolling back or refunding a transaction in progress if `.abort()` is called, however.

ianbjacobs:
Thanks Rouslan. As we discussed at the 23 May call [1], the sense I have is that an event (such as Adrian's proposal) would be a signal from the merchant, and the payment app could behave in a variety of ways (depending on payment method, status of payment, etc.) and inform the site of the outcome.

[1] https://www.w3.org/2017/05/23-apps-minutes#item08

rsolomakhin:
Happy to review any pull requests.

ianbjacobs:
@adrianhopebailie, would you be willing to write up a pull request with your proposal? See notes from the 6 June apps task force call. Thanks!

Ian

rsolomakhin:
I sketched a quick proposal in pull request https://github.com/w3c/payment-handler/pull/170.

adrianhopebailie:
@rsolomakhin beat me to it! Thanks Rouslan

ianbjacobs:
[...]

PR #170, opened 2017-06-06
Add CanMakePaymentEvent.

rsolomakhin:
Please do not merge this pull request yet. I've written this up for reference while we get implementation experience for the proposed `CanMakePaymentEvent` and `AbortPaymentEvent` in Chrome. These events address issues https://github.com/w3c/payment-handler/issues/157 and https://github.com/w3c/payment-handler/issues/117.

All comments are welcome.

***
<a href="https://pr-preview.s3.amazonaws.com/rsolomakhin/payment-handler/pull/170.html" title="Last updated on Mar 26, 2018, 3:29 PM GMT (65e5ec4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-handler/170/bba277f...rsolomakhin:65e5ec4.html" title="Last updated on Mar 26, 2018, 3:29 PM GMT (65e5ec4)">Diff</a>


rsolomakhin:
@gogerald @romandev heads up!

ianbjacobs:
Thank you, @rsolomakhin, for starting this! I have a couple of comments that suggest some updates to the pull request:

- We will need to update the introductory text (Overview of Handling Payment Requests) to indicate at a high level how capability matching happens (through the event). We should also add a mention of AbortPaymentEvent() to the overview.

- I think that the addition of CanMakePaymentEvent means that  PaymentInstrument.capabilities is no longer required and can be deleted. 

- Let's augment examples with these two events. In particular, for CanMakePaymentEvent,
let's include an example of basic card capability matching.

- Similarly, the registration example in 5.7 will need to drop mention of the capabilities method.

- Regarding AbortPaymentEvent(), I think we need some additional explanatory text that the event does not mean that the payment app must abort the payment in order to conform. As we discussed on the call, this is a signal from the merchant and behavior may change depending on payment method and transaction status.

rsolomakhin:
Thank you for the review, @ianbjacobs. I've addressed your comments. My example for `CanMakePaymentEvent` does not have capability matching, because that's highly dependent on how the author decides to structure their payment handler code.

ianbjacobs:
Really nice work, @rsolomakhin! I have some small editorial suggestions (that I will hold onto for now). 

@romandev and @adamroach (yes, you! :) would you like to weigh in?

Ian

ianbjacobs:
@rsolomakhin, another comment: we talked about clarifying that user agents should support disabling the CanMakePaymentEvent. I see this text in the algorithm:

 "If <a>user agent</a> settings prohibit usage of <a>CanMakePaymentEvent</a> (e.g., in private browsing mode), terminate these steps."

It seems we should add a bullet to section 10.1 ( Information about the User Environment ) such as:

   * User agents should allow users to disable support for the CanMakePaymentEvent. 

Thoughts?

Ian

rsolomakhin:
Good idea. Done.

adrianhopebailie:
@rsolomakhin this is great! Thanks.

As an aside this makes the difference between payment methods that use short-string identifiers and those that use URLs smaller too. Since the matching logic can now sit in the handler it's possible for URL-based methods to also have advanced matching logic.

I'm much less grumpy than I was when I wrote https://github.com/w3c/browser-payment-api/pull/536#issuecomment-307813932 😄

index.html
```diff
@@ -1679,6 +1819,129 @@
...
+          Example of handling the <a>AbortPaymentEvent</a>
+        </h2>
+        <p>
+          This example shows how to write a service worker that listens to the
+          <a>AbortPaymentEvent</a>. When a <a>CanMakePaymentEvent</a> is
+          received, the service worker always returns true.
+        </p>
+        <pre class="example js" title="Handling the AbortPaymentEvent">
+          self.addEventListener('abortpayment', function(e) {
+            e.respondWith(true);
```
jnormore: Should we provide any guidance around how long a payment handler should (or will be allowed) to block this request before responding?

rsolomakhin: Good idea. The [PaymentRequest spec](https://w3c.github.io/browser-payment-api/) mentions timeouts. I've added similar text here.

Chrome will probably set the timeout period to 400ms, but the exact amount of time should be at implementer's discretion, in my opinion.

rsolomakhin:
FYI, I've removed `"total"` from `CanMakePaymentEvent` to avoid user fingerprinting.

index.html
```diff
@@ -1251,6 +1244,164 @@
...
+            dictionary CanMakePaymentEventInit : ExtendableEventInit {
+              USVString topLevelOrigin;
+              USVString paymentRequestOrigin;
+              sequence&lt;PaymentMethodData&gt; methodData;
+              sequence&lt;PaymentDetailsModifier&gt; modifiers;
+            };
+          </pre>
+          <p>
+            The <dfn>topLevelOrigin</dfn>, <dfn>paymentRequestOrigin</dfn>,
+            <dfn>methodData</dfn>, <dfn>total</dfn>, and <dfn>modifiers</dfn>
```
romandev: nit: You missed this line. (Please remove |total|) :P

webpayments-specs:
Done.

--Rouslan

index.html
```diff
@@ -1251,6 +1244,163 @@
...
+      </h2>
+      <p>
+        If the <a>payment handler</a> supports <a>CanMakePaymentEvent</a>, the
+        <a>user agent</a> uses it to help with filtering of the available
+        payment handlers.
+      </p>
+      <p>
+        Implementations may impose a timeout for developers to respond to the
+        <a>CanMakePaymentEvent</a>. If the timeout expires, then the
+        implementation will behave as if <dfn>respondWith()</dfn> was called
```
romandev: I think it should be `<a data-lt="CanMakePaymentEvent.respondWith()">respondWith()</a>`. You are defining it below.

index.html
```diff
@@ -1251,6 +1244,163 @@
...
+        };
+        </pre>
+          <p>
+            The <dfn>topLevelOrigin</dfn>, <dfn>paymentRequestOrigin</dfn>,
+            <dfn>methodData</dfn>, and <dfn>modifiers</dfn> members share their
+            definitions with those defined for <a>PaymentRequestEvent</a>.
+          </p>
+        <section>
+          <h2>
+            <dfn>respondWith</dfn>() method
```
romandev: Should be `<dfn>respondWith()</dfn>`.
The `<dfn>respondWith</dfn>` is already defined by IDL. So, this statement can cause duplication definition error.

FYI,
`<a data-lt="CanMakePaymentEvent.respondWith"></a>` makes a link which can jump to IDL definition.
(line 1303 in your patch)

`<a data-lt="CanMakePaymentEvent.respondWith()"></a>` makes a link which can jump here.

index.html
```diff
@@ -1679,6 +1820,135 @@
...
+        handler</a> that is currently handling <a>PaymentRequestEvent</a>.
+        There is no expectation that the <a>payment handler</a> would roll back
+        a transaction or issue a refund. (This is out of scope.) The <a>payment
+        handler</a> should respond with a boolean that indicates whether the
+        abort was successful.
+      </p>
+      <p>
+        Implementations may impose a timeout for developers to respond to the
+        <a>AbortPaymentEvent</a>. If the timeout expires, then the
+        implementation will behave as if <dfn>respondWith()</dfn> was called
```
romandev: Should be `<a data-lt="AbortPaymentEvent.respondWith()">respondWith()</a>`

index.html
```diff
@@ -1679,6 +1820,135 @@
...
+        </p>
+        <pre class="idl">
+        [Constructor(DOMString type, ExtendableEventInit eventInitDict), Exposed=ServiceWorker]
+        interface AbortPaymentEvent : ExtendableEvent {
+          void respondWith(Promise&lt;boolean&gt;paymentAbortedResponse);
+        };
+        </pre>
+        <section>
+          <h2>
+            <dfn>respondWith</dfn>() method
```
romandev: Should be `<dfn>respondWith()</dfn>`.

romandev (commented):
Overall looks good to me! Thank you for this change.
I think there might need a little bit of improvements in order to reduce duplicated parts but we can revisit it later.
I'm also working on implementing this change in Blink. So, I'd like to merge this patch as soon as possible if there is no objection.

I found that there are some respec error and warnings in your patches
Could you fix them if you don't mind?
I left some inline comments.

Even if you fix them, you can still face the following warning.
"No <dfn> for operation respondWith in PaymentRequestEvent."
If you fix the PaymentRequestEvent in the same way, it will work well.

ianbjacobs:
@romandev wrote:

  "I'm also working on implementing this change in Blink. So, I'd like to merge this patch as soon as possible if there is no objection."

That's great news! The idea with this pull request was that we would not merge until we had
some implementation experience, so it's great that you are implementing. I think that once you
and @rsolomakhin are satisfied with the text and the implementation, we should merge. Does that sound ok?

We'll go over this on today's payment apps call.

Ian

rsolomakhin:
I won't be able to attend the meeting this week, btw. Talk to you next week.

On Jul 11, 2017 8:56 AM, "ianbjacobs" <notifications@github.com> wrote:

@romandev <https://github.com/romandev> wrote:

"I'm also working on implementing this change in Blink. So, I'd like to
merge this patch as soon as possible if there is no objection."

That's great news! The idea with this pull request was that we would not
merge until we had
some implementation experience, so it's great that you are implementing. I
think that once you
and @rsolomakhin <https://github.com/rsolomakhin> are satisfied with the
text and the implementation, we should merge. Does that sound ok?

We'll go over this on today's payment apps call.

Ian

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/w3c/payment-handler/pull/170#issuecomment-314434536>,
or mute the thread

index.html
rsolomakhin: Done.

index.html
rsolomakhin: Done.

index.html
rsolomakhin: Done.

index.html
rsolomakhin: Done.

rsolomakhin:
@ianbjacobs @romandev I've fixed up the tidy and respec errors in my pull request. Let's merge.

ianbjacobs:
Thank you, @rsolomakhin.

@romandev and @gogerald, would either (or both) of you like to review and express support of this pull request? Once reviewed I can merge it.

Thank you!

Ian

gogerald:
It looks good to me. 

Just get one idea not sure how useful it is. Would it better to send payment request id to payment handler when aborting payment request? 

It looks not very useful since the id might not unique in the universe and there is supposed to have one payment request showing at a time.
https://www.w3.org/TR/payment-request/#dom-paymentdetailsinit-id

adamroach:
So, to be clear, this looks like every matching registered payment handler will (be able to) get an event whenever a payment is made? So, like, if I'm paying with Basic Card, and I have twelve different payment handlers registered that can handle Basic Card, they're _all_ going to find out that I'm buying something instead of just the one that I use?

That seems like a _huge_ privacy violation. Seriously, you need to get buy-in from the Privacy Interest Group on this.

cc @dbaron -- IMHO, this is a showstopper for implementation in Firefox.

ianbjacobs:
Hi @admaroach!

What changed (at least slightly) from previous discussions about dispatching to "all the handlers" was the introduction of the idea that the user can disable this querying.
That is the purpose of this sentence in the proposal:

   "User agents should allow users to disable support for the CanMakePaymentEvent."

[I hit send too soon, sorry]

Can you indicate whether that makes the proposal more acceptable?
Would "MUST" in that sentence be an improvement?

Thank you for input on this!

Ian

adamroach:
@ianbjacobs -- That seems pretty unsatisfying, as answers go, as it leaks private information by default and requires users to take action (which presumably makes their payment handlers work less effectively?) to protect it. That's backwards from how we design every other privacy-impacting web feature. If this behavior is present, it needs to be opt-in and handled the same as other permissions (like geolocation and camera access). In particular, this means that payment handlers need to be able to work _well_ without the mechanism, which would seem to require keeping the capabilities information in place.

ianbjacobs:
Hi @adamroach,

(For the moment I'll pursue the idea that we can improve this proposal but I remain open to other proposals.)

Key points for discussion (based on your comment):

 * Clearly defining user agent matching behavior when canMakePayment() is disabled.
 * Disabled by default (that is: "opt in"). I think that "opt in" should be "global" 
   rather than "per site".
 * Enabled via standard permissions mechanism
 * I could go either way with "SHOULD" or "MUST" support configuration, but if the
   default mode is "disabled" then without configuration the functionality is meaningless.

Thanks again for the input to the discussion!

Ian

rsolomakhin:
I'm open to seeing counter proposal pull requests.

rsolomakhin:
@adamroach: What about introducing a list of restrictions for `CanMakePayment`? For example, we can fire `CanMakePayment` events only for handlers with URL-based payment method names. Also, we can remove "topLevelOrigin" and "paymentRequestOrigin" from the `CanMakePaymentEvent`. For "basic-card" we can define browser-based filtering algorithm.

marcoscaceres (dismissed):
Needs tests

adrianhopebailie:
@marcoscaceres we are not yet at a level of maturity where tests are required for all changes. This specification is a working draft that reflects a a fast evolving single implementation where we are currently doing a lot of experimentation.

rsolomakhin:
FYI, @ianbjacobs and I have talked to @adamroach, who has stated that Mozilla won't be able to implement the `CanMakePayment` event as described in this pull request. We are sticking with filtering based on `capabilities`.

As for the `AbortPayment` event, that is not contentious, so I will split it off into a separate pull request to merge it.

romandev:
Hi all,

I tried to understand your discussion but I didn't understand fully due to my lack of insight.

If my understanding is correct, the problem is that CanMakePayment event can be triggered with no user interaction. For example, once a user registers a payment handler provided by bobpay.com, calling canMakePayment() on the merchant site will trigger CanMakePayment event to the SW provided by bobpay.com with no user interaction. At that time, bobpay.com might get information that the user has purchased something.

However, I thought that we already have requestPermission() for it. We can add a permission check before CanMakePayment event is triggered. (Rouslan's PR is not represented yet but we can add them.)

Perhaps you might think that a complicated UX flow should be added to check for additional permissions. But we already ask a user before registering a payment handler. So, I think we can recycle existing UX and ask whether allowing CanMakePayment event together.

For example, including a permission for receiving CanMakePayment event.
```
 Do you want to register a new payment handler provided by bobpay.com?
 You need to allow the following permissions:
  - canmakepayment event (blah blah)
  - paymentrequest event (blah blah)
```

In case of no CanMakePayment event,
```
 Do you want to register a new payment handler provided by bobpay.com?
 You need to allow the following permissions:
  - paymentrequest event (blah blah)
```

I think this is very similar to push notification feature. Once a user allows a permission to receive push event for some origin, the site can trigger push events at anytime. So, we can write a polyfill for CanMakePayment event using push notification feature. It might mean that CanMakePayment's privacy risk is not significantly different from PushNotification's privacy risk.

WDYT?

ianbjacobs:
Hi @romandev,

I think you described the privacy concern appropriately: there is concern about (all) payment apps learning about the user's interactions with the merchant BEFORE the user has selected a payment app.

One idea, as you describe, is to get user permission. However, I think that there is some concern that it may be difficult for many users to understand the real implication of the permission.

I forgot to link to yesterday's discussion of this:
 https://www.w3.org/2017/08/29-apps-minutes.html

The two proposals other than canMakePaymentEvent() we are currently investigating are:

 * Static registration of PMI + filters. 
 * Dropping this mechanism entirely, as there is a hypothesis that payment apps may be
   more motivated to be displayed than to be filtered out.

Ian

rsolomakhin:
FYI, I've updated this pull request with a compromise to filter standardized payment methods (e.g., "basic-card") using `capabilities`, but fire `CanMakePaymentEvent` in payment handlers that support only URL-based payment method identifiers (e.g., "https://android.com/pay").

One interesting edge case to consider is when a merchant requests payment via either type of payment method (e.g., both "basic-card" and "https://android.com/pay") and the payment handler also has support for both of these. In this case, I propose that the user agent does not fire the `CanMakePaymentEvent` and behaves as if the payment handler returned "false" for the URL-based payment method. The standardized payment method can still be used in this case.

ianbjacobs:
Thanks for the update, @rsolomakhin. Let's put this on the FTF agenda.

Ian

jenan-stripe:
@rsolomakhin Is it correct that this algorithm would prohibit Chrome's current implementation of `basic-card` treating `canMakePayment` like Apple's `canMakePaymentsWithActiveCard`? Would it also mean that any registered `basic-card` handler would cause `(new PaymentRequest({ ... 'basic-card' ... })).canMakePayment()` to resolve to `true`, regardless of active/usable card?

rsolomakhin:
@jenan-stripe That was not my intention. Can you elaborate more on how you envision this happening?

I expect a payment handler to provide the list of all of its active cards to the browser like so:

```javascript
navigator.serviceWorker.register('/pw/app.js').then(function(registration) {
  registration.paymentManager.userHint = '(Visa ****1111)';
  registration.paymentManager.instruments.set('0', {
    name: "Visa ****1111",
    icons: [{
      src: '/pay/visa.png',
      sizes: '32x32',
      type: 'image/png'
    }],
    enabledMethods: ['basic-card'],
    capabilities: {
      supportedNetworks: ['visa'],
      supportedTypes: ['credit']
    }
  });
});
```

Then `new PaymentRequest([{supportedMethods: 'basic-card'}], shoppingCart).canMakePayment()` should return true because there's an active card in the payment handler. Note that `new PaymentRequest([{supportedMethods: 'basic-card', data: {supportedTypes: ['debit']}}], shoppingCart).canMakePayment()` would return false because of mismatch in `supportedTypes` in this example.

By the way, the current implementation of `basic-card` in Chrome via autofill cards stored on disk is not affected by this payment handler specification.

jenan-stripe:
@rsolomakhin Ah! That probably covers it then. 👍

rsolomakhin:
@ianbjacobs: I've verified that I'm happy with this pull request and ready to discuss it at our next meeting. I believe @marcoscaceres would want to see a pull request into web platform tests first before merging this pull request. Is that correct, @marcoscaceres?

rsolomakhin:
@marcoscaceres: I've written a couple of tests for `CanMakePaymentEvent` in pull request https://github.com/w3c/web-platform-tests/pull/8377.

index.html
```diff
@@ -913,6 +918,85 @@ <h2>
...
+            <ol>
+              <li>For each <var>key</var> in <var>methodData</var>:
+                <ol>
+                  <li>If the intersection of <var>methodData[key]</var> and
+                  <var>instrument.capabilities[key]</var> is empty, return
+                  "false".
+                  </li>
+                </ol>
+              </li>
+              <li>Otherwise, return "true".
```
marcoscaceres: please remove the quotes from "true" and "false", otherwise, people might get confused that they should be expecting a string.

index.html
```diff
@@ -913,6 +918,85 @@ <h2>
...
+        <section id="capabilities-example" class="informative">
+          <h2>
+            How to specify capabilities
+          </h2>
+          <p>
+            Example of how a payment handler should provide the list of all its
+            active cards to the browser.
+          </p>
+          <pre class="example js" title="Specifying capabilities">
+navigator.serviceWorker.register('/pw/app.js').then(function() {
```
marcoscaceres: Nit: we should rewrite this using async/await. It will make it much easier to read. Also, don't be shy about indenting this to match the rest of the content.

marcoscaceres:
> Is that correct, @marcoscaceres?

That's correct. Thanks for starting to put the tests together. I'll give this a more thorough review as we work on the tests.

index.html
rsolomakhin: Done.

index.html
rsolomakhin: Done.

rsolomakhin (commented):
@marcoscaceres: Please take another look.

marcoscaceres:
At a conf today, so will try to review tomorrow or Friday.

index.html
```diff
@@ -368,7 +373,7 @@ <h2>
       interface PaymentManager {
         [SameObject] readonly attribute PaymentInstruments instruments;
         [Exposed=Window] static Promise&lt;PermissionState&gt; requestPermission();
-        attribute DOMString userHint;     
+        attribute DOMString userHint;
```
marcoscaceres, line 322: I'm scared of this... seems easy for developers to screw up and expose too much information, or in an ugly way "****". I'd rather the UA be in charge of masking instrument information.

rsolomakhin:
The tests landed in https://github.com/w3c/web-platform-tests/pull/8680. This is ready to land as well now, IMHO.

rsolomakhin (commented):
@marcoscaceres, @ianbjacobs: Please take a look. I've resolved the merge conflicts.

index.html
```diff
@@ -847,6 +852,163 @@ <h2>
         </section>
       </section>
     </section>
+    <section id="canmakepayment">
+      <h2>
+        Can make payment
+      </h2>
+      <p>
```
marcoscaceres, line 859: We should add a `<div class="issue" data-number>` here describing the privacy concern. If we can dig up Adam's original comment, maybe we can copy/paste text from there.

index.html
rsolomakhin: @marcoscaceres: One way we originally envisioned doing what you propose is providing a way for the payment handler to specify which payment instrument is the default, so the UA can use its description as the user hint when not showing all instruments. However, this confused pretty much everyone, because people thought the default payment handler is the one that's currently selected inside of the payment handler and that had some kind of impact on processing the `PaymentRequestEvent`, which was not the intention. Our intention is to provide some user-specific data next to the common name of the payment handler, because we have seen higher checkout completion rates when we did this for shipping addresses. This user-specific data does not have to be the last four digits of the card number. It can be user's email address or username, for example, which is not something that you would typically find in the description of their instruments. Hence, we believe that giving the web developers the power to explicitly specify the user hint is the best approach that we could come up with. Do you have anything better in mind?

index.html
rsolomakhin: @marcoscaceres : This is a bad merge resolution on my part. I've restored the issue marker.

rsolomakhin:
@marcoscaceres: Sorry about the bad merge! Please take another look. The patch has the issue marker now.

By the way, this patch does not alter the `userHint` parameter. It shows up in the diff because I removed trailing whitespace from it.

index.html
marcoscaceres: > Do you have anything better in mind?

Was thinking, if you know the instrument type (e.g., this is an ISO-7812 conforming card number), the UA can mask it as it would normally for credit cards.

For other types, I'm not sure... but maybe there are conventions already.

ianbjacobs: Hi Marcos,

We also heard from payment handler people that users might have the ability to label instruments themselves to help improve recognition. For example, "Business Account" or "Trust account". 

Therefore, by design this was intentionally open to a variety of strings.

Ian

rsolomakhin: @marcoscaceres: The UA does not have the raw instrument data (i.e., card number), only its name, icon, and capabilities. Therefore, it's not possible for UA to mask the raw instrument data. Masking the name would seem strange.

index.html
```diff
@@ -847,6 +852,248 @@ <h2>
...
+              </li>
+              <li>Otherwise, return <code>true</code>.
+              </li>
+            </ol>
+          </li>
+          <li>Otherwise, fire the <a>CanMakePaymentEvent</a> in the payment
+          handler and return the result.
+          </li>
+        </ol>
+        <div class="issue" data-number="157">
```
marcoscaceres, line 1051: Should we link to the wiki: https://github.com/w3c/payment-request-info/wiki/PaymentMethodPractice#canmakepaymentevent-implementation

index.html
marcoscaceres: Let's come back to this later

ianbjacobs: I've created a separate issue:
 https://github.com/w3c/payment-handler/issues/273

index.html
rsolomakhin: Done. Please take a look, @marcoscaceres.

marcoscaceres (approved):

marcoscaceres:
@rsolomakhin, just wondering, what ever happened to AbortPaymentEvent... is it still in Chrome?  https://bugs.chromium.org/p/chromium/issues/detail?id=736745 

This doesn't appear to be in the spec. cc @a2sheppy. The MDN folks are working on updating docs around payment handlers...

rsolomakhin:
@marcoscaceres : AbortPaymentEvent appears to be in Chrome, but I believe @romandev and @gogerald  implemented it based on a pull request into the spec, if memory serves me right. According to the comment on issue https://github.com/w3c/payment-handler/issues/117#issuecomment-422851163, the pull request needed to be rewritten. I will be doing that soon.

rsolomakhin:
@a2sheppy: It's probably best to treat AbortPaymentEvent as unimplemented until the it's in the spec.

a2sheppy:
@rsolomakhin: OK, I will just ignore it. Fortunately, that's easy because it wasn't until a bit later on that I realized this was Payment Handler, not Payment Request, and we're not currently documenting PH. :)

Issue #169, opened 2017-06-05
normative "A single payment handler should not be allowed to open more than one client window using this method."


gogerald:
There is a non-normative note that "A single payment handler should not be allowed to open more than one client window using this method."

Should we normalize it?

ianbjacobs:
@romandev, this seems related to #163. It makes sense to move the constraint to the open window algorithm (and out of the Note). Thank you!

Ian

romandev:
> @romandev, this seems related to #163. It makes sense to move the constraint to the open window algorithm (and out of the Note). Thank you!

IMHO, the spec algorithm is already explaining the problem.

> If event.[[windowClient]] is not null, then:
> > If event.[[windowClient]].visibilityState is not "unloaded", reject promise with a DOMException whose name is "InvalidStateError" and abort these steps.

When openWindow is called at first time, the [[windowClient]] will be null. But when it is called at the next time, it will not be null and then if the windowClient is *not* unloaded, then, reject InvalidStateError.
In other words, if window is open, then reject promise with InvalidStateError.

But I think it's better to check it before navigation happens.

rsolomakhin:
Can we close as "working as intended"? @romandev

romandev:
@rsolomakhin I think we can close this issue :-)

Issue #168, opened 2017-06-05
is PaymentRequestEvent.openWindow should check origin before navigation instead of after navigation


gogerald:
OpenWindow algorithm 8-4 checks origin after navigation, however from this bug discussion https://github.com/w3c/payment-handler/issues/163, it looks we would like to check origin before navigation.

ianbjacobs:
@romandev, this seems related to https://github.com/w3c/payment-handler/issues/163
so please note this in any updates to the algorithm. Thanks!

Ian

romandev:
IMHO, even if we add this check to the algorithm, we should keep the original check logic.
For example, the input origin was a proper origin, but if the origin is set as unexpected origin after navigation(e.g. redirection), then the SW might not controll the windowClient anymore.

ianbjacobs:
Closed (I believe) through PR #171

Issue #163, opened 2017-05-25
Missing question mark,


gogerald:
is 
```
[Exposed=ServiceWorker]
interface PaymentRequestEvent : ExtendableEvent {
    ...
    Promise<WindowClient> openWindow(USVString url);
   ...
};
```
should be
```
[Exposed=ServiceWorker]
interface PaymentRequestEvent : ExtendableEvent {
    ...
    Promise<WindowClient?> openWindow(USVString url);
   ...
};
```
to allow return null

romandev:
According to the OpenWindow Algorithm 8-4, unlike clients.openWindow(), this function will create controlled window client only.
So, it will always be non-null?

[1] https://w3c.github.io/payment-handler/#dfn-open-window-algorithm

rsolomakhin:
@romandev : What does the word "controlled" mean in this context?

@gogerald : In what circumstances would `openWindow()` return a null?

gogerald:
The OpenWindow Algorithm 8-6 checks "If event.[[windowClient]] is not null", does that mean it could be null?

Right now, I can think of the situation is that, the user clicks 'pay' button, navigates to the other windows, then we might not want to open the new window.

rsolomakhin:
Maybe `Promise<WindowClient>` is referred to in in OpenWindow Algorithm 8-5: "Let client be the result of running the create window client algorithm with newContext as the argument."

IMHO, if `e.openWindow()` fails, it should reject the promise instead of returning a null. @marcoscaceres : Thoughts?

romandev:
@rsolomakhin,

If a SW client is controlled by an active worker, we can says that the SW client is a controlled client.
(FYI, The SW client can be window client, worker and so on.)

For example, let's imagine that the origin of SW is "https://example.com".
In the SW context,
```
clients.openWindow("https://example.com/abc").then(windowClient => {
    // windowClient is controlled.
});
```

```
clients.openWindow("https://uncontrolled.com").then(windowClient => {
    // this windowClient is not controlled. So, should be null.
});
```

rsolomakhin:
Thank you for the explanation. That makes a lot of sense. IMHO, the payment handler SW should be allowed to open only controlled clients. If a client is not controlled, then `e.openWindow()` promise should be rejected with `SecurityError`. I believe that's already in the spec, right?

rsolomakhin:
Yep:
> If the origin of newContext is not the same as the service worker client origin associated with the payment handler, then:
[...]

PR #171, opened 2017-06-06
Improve open window algorithm.

romandev:
This change is fixing the following things:
  - Early reject SecurityError if the input url's origin is not the same as the
    service worker's origin associated with the payment handler.
  - Check whether already opened window exists before navigation.

This change is related to #163, #168, #169.


romandev:

rsolomakhin (approved):
LGTM

Issue #125, opened 2017-04-04
Image object is undefined


ianbjacobs:
The "icons" field states:

 "The icons member is an array of image objects..."

but "image objects" are undefined here. The spec is clearly referencing the Web App Manifest term:
 https://www.w3.org/TR/appmanifest/#dfn-image-object

There is a lengthy section of that spec devoted to image objects. Should we:

 * Include a normative reference to that spec, or
 * Make use of a simplified data structure for our needs here

(I don't have an informed opinion yet.)

We've raised this issue after discussion at the 4 April 2017 payment apps task force call [1].

Ian

[1] https://www.w3.org/2017/04/04-apps-minutes#item07

rsolomakhin:
Let's include a normative reference the algorithms for processing ImageObject. However, the manifest spec does not have WebIDL, so we would need to define it in our own spec.

marcoscaceres:
We should probably start defining things as IDL dictionaries in the manifest spec - or else it's gonna get messy if other specs want to use this stuff.  It's ok if we work out the details here and then upstream what we come up with.

ianbjacobs:
I believe this is closed with this PR merged:
 https://github.com/w3c/payment-handler/pull/174

Ian

PR #174, opened 2017-06-08
Add ImageObject dictionary.

romandev:
The PaymentInstrument and the PaymentWallet have a sequence of ImageObject
as member, but there is no definition of ImageObject in the current spec.
We already know implicitly that is is similar to the image object in the
manifest spec[1]. But the does not have WebIDL definition of it, so it is
difficult to represent our data structure. To resolve the issue, this change
is adding ImageObject dictionary in WebIDL level.

### DIFF
--- a/index.html
+++ b/index.html
@@ -695,10 +695,52 @@
             When called, this method executes the following steps:
           </p>
           <ol>
+            <li>If the <a data-lt="PaymentInstrument.icons">icons</a> member of
+            <var>details</var> is present, then:
+              <ol>
+                <li>Let <var>convertedIcons</var> be the result of running the
+                <a>convert image objects</a> algorithm passing
+                <var>details</var>.<a data-lt=
+                "PaymentInstrument.icons">icons</a> as the argument.
+                </li>
+                <li>If the <var>convertedIcons</var> is an empty
+                <a>Sequence</a>, then return a <a>Promise</a> rejected with a
+                <a>TypeError</a>.
+                </li>
+                <li>Set <var>details</var>.<a data-lt=
+                "PaymentInstrument.icons">icons</a> to
+                <var>convertedIcons</var>.
+                </li>
+              </ol>
+            </li>
             <li>Let <var>p</var> be a new promise.
             </li>
             <li>Run the following steps in parallel:
               <ol>
+                <li>If the <a data-lt="PaymentInstrument.icons">icons</a>
+                member of <var>details</var> is present, then for each
+                <var>icon</var> in <var>details</var>.<a data-lt=
+                "PaymentInstrument.icons">icons</a>:
+                  <ol>
+                    <li>If the user agent wants to display the <var>icon</var>,
+                    then:
+                      <ol>
+                        <li>Let <var>fetchedImage</var> be the result of
+                        running the <a data-cite=
+                        "!appmanifest#fetching-image-objects">fetching image
+                        object</a> passing <var>icon</var> as the argument.
+                        </li>
+                        <li>Set <var>icon</var>.<a>[[\fetchedImage]]</a> to
+                        <var>fetchedImage</var>.
+                        </li>
+                      </ol>
+                    </li>
+                    <li>Otherwise, if no images are fetched, the user agent may
+                    set <var>icon</var>.<a>[[\fetchedImage]]</a> to fallback
+                    image.
+                    </li>
+                  </ol>
+                </li>
                 <li>If the collection contains a <a>PaymentInstrument</a> with
                 a matching <var>instrumentKey</var>, replace it with the
                 <a>PaymentInstrument</a> in <var>details</var>.
@@ -746,7 +788,7 @@
         <pre class="idl">
       dictionary PaymentInstrument {
         required DOMString name;
-        sequence&lt;ImageObjects&gt; icons;
+        sequence&lt;ImageObject&gt; icons;
         sequence&lt;DOMString&gt; enabledMethods;
         object capabilities;
       };
@@ -788,11 +830,6 @@
             <a>supportedTypes</a>.
           </dd>
         </dl>
-        <p class="issue" title="ImageObjects not Defined" data-number="125">
-          ImageObjects comes from the Web App Manifest specification. Should we
-          reference the definition normatively, or make use of a simpler
-          structure here?
-        </p>
       </section>
       <section data-dfn-for="PaymentWallets" data-link-for="PaymentWallets">
         <h2>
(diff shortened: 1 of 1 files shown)


romandev:

There has been no definition until now. So, it hasn't been easy to implement the icon fetcher in browser side.

ianbjacobs:
This relates to issue 125
https://github.com/w3c/payment-handler/issues/125

I have been wondering how much of manifest to pull in. It looks like you've taken the approach of having a bit inline, but also references to manifest (for fetching) and HTML (for some aspects of size information). I'd love to hear from other developers familiar with those specs if you like this balance.

Per Marcos' suggestion, @rsolomakhin (or another implementer), please review. Thanks!
Ian



Ian

index.html
```diff
@@ -930,6 +966,16 @@
                 <var>details</var> as a new member of the collection and
                 associate it with the key <var>walletKey</var>.
                 </li>
+                <li>If the <a data-lt="PaymentWallet.icons">icons</a> member of
+                <var>details</var> is present, then for each <var>icon</var> in
```
rsolomakhin: We should download the best icon for our use case, not all of them :-)

index.html
```diff
@@ -695,6 +695,21 @@
...
+                is not a <a data-cite="#valid-mime-type">valid MIME type</a> or
+                the value of type is not a supported media format, then return
+                a <a>Promise</a> rejected with a <a>TypeError</a>.
+                </li>
+                <li>If <var>icon</var>.<a data-lt="ImageObject.sizes">sizes</a>
+                is not a <a data-lt="ImageObject.sizes">valid value</a>, then
+                return a <a>Promise</a> rejected with a <a>TypeError</a>.
+                </li>
+              </ol>
+            </li>
```
rsolomakhin: Let's require the URL of the icon to be from a secure origin. (SecureContext)

rsolomakhin: I suggest we prohibit relative URLs, because it's not clear how to resolve these into absolute URLs in the context of a service worker.

rsolomakhin (changes requested):
Great work! A few comments inline.

index.html
romandev: IMHO, we can use relative URLs.

I added parsing step for it as follows:

>  parsing image.src with the context object's relevant settings object's API base URL.

romandev:
@rsolomakhin, @ianbjacobs Thank you for review.
I addressed your comments. PTAL

index.html
rsolomakhin: What is "the context object's relevant settings object's API base URL"? Is the the "scope" of the service worker or the path of the service worker's JavaScript file or something else entirely?

romandev: @rsolomakhin,

In a case of SW, it will be script_url.
If the script_url is https://example.com/abcd/sw.js, then:
```
// SW context
clients.openWindow('/test1.html');      // https://example.com/test1.html
clients.openWindow('../test2.html');    // https://example.com/test2.html
clients.openWindow('./test3.html');     // https://example.com/abcd/test3.html
```

In a case of document, it will be document_url.
If the document_url is https://example.com/hello/world/test.html, then:
```
// Window global context
navigator.serviceWorker.register('/test1.js');     // https://example.com/test1.js
navigator.serviceWorker.register('../test2.js');   // https://example.com/hello/test2.js
navigator.serviceWorker.register('./test3.js');    // https://example.com/hello/world/test3.js
```

The existing SW APIs already uses the processing. [1][2]

[1] https://w3c.github.io/ServiceWorker/#clients-openwindow
[2] https://w3c.github.io/ServiceWorker/#navigator-service-worker-register

ianbjacobs: @romandev, thank you for the examples. One question: given this specification's reliance on service workers, is that the only context that will be relevant here? And if so, should we say that clearly in the bullet that talks about the origin of the base URL?

Ian

romandev: @ianbjacobs,

Sorry for delayed reply. I missed this comment.
IMHO, the following parsing step is enough to explain how the relative url should be resolved clearly.
 
> parsing image.src with the context object's relevant settings object's API base URL.
 
The ImageObject is used during PaymentInstrument.set().
BTW, the PaymentInstrument is an extension of ServiceWorkerRegistration.
So, we can access it in both of window global context and worker context but only need to consider which context the API is running on.

**"the context object's relevant settings object's API base URL"**

It means that..
If the top level window, then the base url is window's location url without filename.
If it is an iframe, then the base url is iframe's url without filename.
If it is a worker, then the base url is script_url without filename.

For example, we can run PaymentInstrument.set() in window global context.
```
<script>
// https://www.example.com/abcd/test.html
navigator.seviceWorker.register(...)
    ...
    .then(registration) {
        registration.paymentManager.paymentInstrument.set({
            name: "Test Pay",
            icons: {
                src: '../icon.png'    /* https://www.example.com/icon.png */
            }
        });
    }
</script>
```

If its context is iframe, no problems.
```
<!-- https://www.example.com/abcd/efgh/index.html -->
<iframe src="https://www.example.com/abcd/efgh/iframe.html"></iframe>
```
```
<script>
// This is an iframe: https://www.example.com/abcd/efgh/iframe.html
navigator.seviceWorker.register(...)
    ...
    .then(registration) {
        registration.paymentManager.paymentInstrument.set({
            name: "Test Pay",
            icons: {
                src: '../icon.png'    /* https://www.example.com/abcd/icon.png */
            }
        });
    }
</script>
```

In a case of worker context,
```
// https://www.example.com/abcd/efgh/sw/pay.js
registration.paymentManager.paymentInstrument.set({
    name: "Test Pay",
    icons: {
        src: '../icon.png'    /* https://www.example.com/abcd/efgh/icon.png */
    }
});
```

ianbjacobs: Thanks very much, @romandev. What do you think of adding those examples after the algorithm in the section "Convert image objects" to show how relative URL conversion would happen in different contexts?

In other words: the normative text may well suffice, but the examples could help explain expected behavior in different scenarios.

romandev: @ianbjacobs Thank you for guide. I added two examples. Could you check my update?

ianbjacobs:
@rsolomakhin, my most recent suggestions are editorial. Could you please review the full PR (including example) and let us know if it looks ok. Seems like we are close to being able to merge the PR. Thanks!

Ian

romandev:
FYI, I've just fixed @ianbjacobs's comments in the latest patch. :-)
Thanks!

index.html
```diff
@@ -695,10 +695,52 @@
...
+                        </li>
+                        <li>Set <var>icon</var>.<a>[[\fetchedImage]]</a> to
+                        <var>fetchedImage</var>.
+                        </li>
+                      </ol>
+                    </li>
+                    <li>Otherwise, if no images are fetched, the user agent may
+                    set <var>icon</var>.<a>[[\fetchedImage]]</a> to fallback
+                    image.
+                    </li>
```
rsolomakhin: Nit: remove this UX guidance. It's not important for compatibility.

index.html
```diff
@@ -918,10 +955,50 @@
...
+                        object</a> passing <var>icon</var> as the argument.
+                        </li>
+                        <li>Set <var>icon</var>.<a>[[\fetchedImage]]</a> to
+                        <var>fetchedImage</var>.
+                        </li>
+                      </ol>
+                    </li>
+                    <li>Otherwise, if no images are fetched, the user agent may
+                    set <var>icon</var>.<a>[[\fetchedImage]]</a> to fallback
+                    image.
```
rsolomakhin: Ditto.

index.html
```diff
@@ -1008,6 +1085,167 @@
...
+          </dd>
+        </dl>
+      </section>
+      <section>
+        <h2>
+          <dfn>Convert image objects</dfn>
+        </h2>
+        <p>
+          When this algorithm with <var>inputImages</var> parameter is invoked,
+          the user agent must runs the following steps:
```
rsolomakhin: s/runs/run

index.html
```diff
@@ -1008,6 +1085,167 @@
...
+              <var>image</var>.<a data-lt="ImageObject.src">src</a> with the
+              <a data-cite="!HTML#context-object">context object</a>'s
+              <a data-cite="!HTML#relevant-settings-object">relevant settings
+              object</a>'s <a data-cite="!HTML#api-base-url">API base URL</a>.
+              </li>
+              <li>If <var>url</var> is failure, then return an empty
+              <a>Sequence</a> of <a>ImageObject</a>.
+              </li>
+              <li>If <var>url</var>'s <a data-cite="!HTML#concept-url-scheme">
+                scheme</a> is not one of "http" and "https", then return an
```
rsolomakhin: Let's allow only HTTPS. Using HTTP is recipe for trouble in payments world.

rsolomakhin (changes requested):
Please prohibit usage of HTTP scheme for icon storage.

rsolomakhin (approved):

Issue #175, opened 2017-06-08
PaymentRequestEvent.total


rsolomakhin:
The [`PaymentRequestEvent.total`](https://w3c.github.io/payment-handler/#dom-paymentrequestevent) is defined to have a type of [`PaymenItem`](https://w3c.github.io/browser-payment-api/#paymentitem-dictionary), which is a dictionary. However, [WebIDL spec](https://heycam.github.io/webidl/#idl-dictionaries) says that attributes should not be dictionaries.

romandev:
I'm not sure but the spec says that:

> The type of the attribute, after resolving typedefs, must not be a nullable or non-nullable version of any of the following types:
> 
> - a sequence type
> - a dictionary type  <<-- Maybe.. possible?
> - a record type
> - a union type that has a nullable or non-nullable sequence type, dictionary, or record as one of its flattened member types

If I'm missing something, please correct me.

https://heycam.github.io/webidl/#idl-attributes

rakuco:
> must not be a nullable or non-nullable version of any of the following types:
> * a dictionary type

Doesn't that corroborate @rsolomakhin's comment? That also matches https://heycam.github.io/webidl/#idl-dictionaries, which says _"Dictionaries must not be used as the type of an attribute or constant."_

ianbjacobs:
Hi all,

Here's some guidance from a draft WebIDL cookbook [1]:

===
Use type any or object and describe in prose that the attribute takes the type of the dictionary.

Example 22
   interface HasRightUsage {
    readonly attribute any dict;
    };

 In your spec, write "dict is of type IAmDictionary dictionary as defined in [REFERENCE]". 
=====

Ian

[1] http://darobin.github.io/api-design-cookbook/#attribute-constant-or-exception-fields

PR #177, opened 2017-06-11
The PaymentRequestEvent.total attribute should be |object| type.

romandev:
The |total| attribute is currently defined as PaymentItem dictionary but
WebIDL spec says that attributes should not be a dictionary[1].
This change is fixing #175 issue.

[1] https://heycam.github.io/webidl/#idl-dictionaries

### DIFF
--- a/index.html
+++ b/index.html
@@ -1314,7 +1314,7 @@
           readonly attribute USVString paymentRequestOrigin;
           readonly attribute DOMString paymentRequestId;
           readonly attribute FrozenArray&lt;PaymentMethodData&gt; methodData;
-          readonly attribute PaymentItem total;
+          readonly attribute any total;
           readonly attribute FrozenArray&lt;PaymentDetailsModifier&gt; modifiers;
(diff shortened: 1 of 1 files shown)


romandev:

index.html
```diff
@@ -1314,7 +1314,7 @@
           readonly attribute USVString paymentRequestOrigin;
           readonly attribute DOMString paymentRequestId;
           readonly attribute FrozenArray&lt;PaymentMethodData&gt; methodData;
-          readonly attribute PaymentItem total;
+          readonly attribute any total;
```
rsolomakhin: Can we use `object` type to match `data` attribute in `PaymentMethodData`?

https://w3c.github.io/browser-payment-api/#paymentmethoddata-dictionary

rsolomakhin (changes requested):

rsolomakhin (approved):

index.html
romandev: Done! Thanks.

Issue #94, opened 2017-01-23
Add explicit permission call to allow payment app to handle payments


adamroach:
In the current document, the request for permission to install an app occurs implicitly as part of the "setManifest()" processing. This is similar to the way permissions are handled in a host of existing web technologies, such as geolocation and getUserMedia.

The more recent pattern for these kinds of permissions requests (e.g., for push notifications) involves an explicit call to request permission before the action that requires permission is performed.

In @marcoscaceres' payment-request-handler doc, he proposes that the `paymentAppManager` object include an explicit `requestPermission()` method rather than this being an implicit step in setting up a payment app. I think this is an improvement, and would support adding it to the payment-app specification.

jakearchibald:
A pattern more like https://github.com/marcoscaceres/payment-request-handler/issues/2 would be in line with other service worker APIs.

marcoscaceres:
Ok, so considering all feedback thus far, I want to avoid asking for permissions on multiple instances (because it leads to redundant code, like):

```JS
// Only the first one matters... so second line is just code duplication.
Promise.all([
  sw1.paymentManager.requestPermission(),
  sw2.paymentManager.requestPermission(),
]).then()
```

Also, I don't know if the app should be allowed to revoke its permission (through some kind on `.unregister()`). That's been [really controversial](https://github.com/w3c/permissions/issues/46), and more of less rejected in permission API discussions (I had to pref that off in Gecko). 

So how about... 

### Permission can only be triggered via user interaction:

![screenshot 2017-01-27 17 20 55](https://cloud.githubusercontent.com/assets/870154/22362255/6c7daa20-e4b5-11e6-8325-73b2f0296693.png)

The user MUST activate the element, in this case a button.

### Getting permission
So, "Let's get permission"... but this this time, using `PaymentManager.requestPermission()` - similarly how Web Notifications does things.

```JS
button.addEventListener("click", async () => {
  let permCheck;
  try {
    permCheck = await navigator.permissions.query({ name: "payments" });
  } catch (err) {
    return; // not supported
  }
  const canProceed = await permissionsCheck(permCheck);
  if (!canProceed) {
    return; // we got denied.
  }
  weCanDoStuff(); // Yay!
}, { once: true });

async function permissionsCheck({ state }) {
  switch (state) {
    case "prompt":
      const result = await PaymentManager.requestPermission();
      return await permissionsCheck(Promise.resolve({ state: result }));
    case "granted":
      return true;
    default:
[...]

PR #180, opened 2017-06-21
Define requestPermission() behavior.

romandev:
In the current spec, there is no definition of the requestPermission() method.
This change is fixing #94 issue.

### DIFF
--- a/index.html
+++ b/index.html
@@ -275,6 +275,9 @@
             "!WEBIDL-LS#invalidstateerror">InvalidStateError</dfn></code>"
             </li>
             <li>"<code><dfn data-cite=
+            "!WEBIDL-LS#notallowederror">NotAllowedError</dfn></code>"
+            </li>
+            <li>"<code><dfn data-cite=
             "!WEBIDL-LS#notfounderror">NotFoundError</dfn></code>"
             </li>
             <li>"<code><dfn data-cite=
@@ -983,7 +986,7 @@
             <li>Let <var>permission</var> be the result of running
               <a data-cite="!permissions#dfn-retrieve-the-permission-state">retrieve
               the permission state algorithm</a> of the permission associated
-              with <a>payment handler</a>.
+              with <a>payment handler</a>'s <a>origin</a>.
             </li>
             <li>If <var>permission</var> is not "granted", then return a
             <a>Promise</a> rejected with a <a>NotAllowedError</a>.
--- a/index.html
+++ b/index.html
@@ -565,6 +565,10 @@
           <h2>
             <dfn>requestPermission()</dfn> method
           </h2>
+          <p class="note">
+            The user agent is NOT REQUIRED to prompt the user to grant
+            permission to the origin for each new supported payment method.
+          </p>
           <p>
             When called, this method executes the following steps:
           </p>
--- a/index.html
+++ b/index.html
@@ -517,7 +517,7 @@
       interface PaymentManager {
         [SameObject] readonly attribute PaymentInstruments instruments;
         [SameObject] readonly attribute PaymentWallets wallets;
-        [Exposed=Window] static Promise&lt;PermisionState&gt; requestPermission();
+        [Exposed=Window] static Promise&lt;PermissionState&gt; requestPermission();
       };
       </pre>
         <p>
@@ -573,7 +573,7 @@
                 <li>Let <var>permission</var> be the result of running
                 <a data-cite="!permissions#dfn-retrieve-the-permission-state">
                   retrieve the permission state algorithm</a> of the permission
-                  associated with <a>payment handler</a>.
+                  associated with <a>payment handler</a>'s <a>origin</a>.
                 </li>
                 <li>If <var>permission</var> is "prompt", ask the user whether
                 allowing adding new payment instruments for the <a data-cite=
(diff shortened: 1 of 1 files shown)


romandev:

Thank you.

index.html
```diff
@@ -513,10 +513,11 @@
           <dfn>PaymentManager</dfn> interface
         </h2>
         <pre class="idl">
+      [SecureContext]
       interface PaymentManager {
         [SameObject] readonly attribute PaymentInstruments instruments;
         [SameObject] readonly attribute PaymentWallets wallets;
-                              Promise&lt;boolean&gt;   requestPermission();
+        [Exposed=Window] static Promise&lt;PermisionState&gt; requestPermission();
```
rsolomakhin: What does  `[Exposed=Window]` do?

index.html
```diff
@@ -561,14 +562,32 @@
...
+                <li>Let <var>permission</var> be the result of running
+                <a data-cite="!permissions#dfn-retrieve-the-permission-state">
+                  retrieve the permission state algorithm</a> of the permission
+                  associated with <a>payment handler</a>.
+                </li>
+                <li>If <var>permission</var> is "prompt", ask the user whether
+                allowing adding new payment instruments for the <a data-cite=
+                "!HTML#current-settings-object">current settings object</a>'s
+                origin is acceptable. If it is, set <var>permission</var> to
+                "granted", and "denied" otherwise.
```
rsolomakhin, line 589: Your [implementation](https://codereview.chromium.org/2946013002/diff/1/third_party/WebKit/Source/modules/payments/PaymentManager.cpp) can also resolve `p` with `"default"`. When would that happen?

index.html
```diff
@@ -697,6 +716,14 @@
             When called, this method executes the following steps:
           </p>
           <ol>
+            <li>Let <var>permission</var> be the result of running
+              <a data-cite="!permissions#dfn-retrieve-the-permission-state">retrieve
+              the permission state algorithm</a> of the permission associated
+              with <a>payment handler</a>.
+            </li>
+            <li>If <var>permission</var> is not "granted", then return a
+            <a>Promise</a> rejected with a <a>SecurityError</a>.
```
rsolomakhin: `NotAllowedError` might be better, IMHO.

index.html
```diff
@@ -953,6 +980,14 @@
...
```
rsolomakhin: Ditto

rsolomakhin (changes requested):
Thank you for the patch! Good work!

index.html
dlongley: I believe that the `[SecureContext]` extended attribute applies to the entire interface (exposing it only in secure contexts) and the `[Exposed=Window]` is a further restriction, specifically, on `requestPermission` to cause it to only be exposed when the global context is `Window` (as opposed to in a `Worker`, for example -- where it won't be accessible given this directive).

dlongley: Also, there's a typo here: `PermisionState` => `PermissionState`

index.html
```diff
@@ -561,14 +562,32 @@
...
           </p>
+          <ol>
+            <li>Let <var>p</var> be a new promise.
+            </li>
+            <li>Run the following steps in parallel:
+              <ol>
+                <li>Let <var>permission</var> be the result of running
+                <a data-cite="!permissions#dfn-retrieve-the-permission-state">
+                  retrieve the permission state algorithm</a> of the permission
+                  associated with <a>payment handler</a>.
```
dlongley: Should this say for the permission associated with `<a>payment handler's</a> <a>origin</a>`?

index.html
romandev: Done.

FYI, I found a history of Notification.requestPermission().
https://lists.w3.org/Archives/Public/public-whatwg-archive/2014Jul/0178.html

index.html
romandev: Done.

index.html
romandev: Done.

index.html
romandev: Done.

index.html
romandev: In the implementation, I defined PaymentHandlerStatus as 'default' instead of 'prompt'.
Because the Notification spec[1] is doing like that.
But I awared that we can use PermissionState in Permission API spec[2] instead of our own definition.

[1] https://notifications.spec.whatwg.org/#enumdef-notificationpermission
[2] https://www.w3.org/TR/permissions/#idl-def-PermissionState

romandev (commented):
@rsolomakhin, @dlongley 

Thank you for review 👍 

I addressed your comments.

index.html
ianbjacobs: Thanks @romandev, @rsolomakhin, and @dlongley!

@romandev, you deleted:
  " The user agent is NOT REQUIRED to prompt the user to grant  permission to the origin for each new supported payment method."

I think we still need that text.

romandev: @ianbjacobs, You're correct!

Done. Thanks.

ianbjacobs:
Thanks @romandev! @rsolomakhin if you are satisfied your comments have been addressed, please give the green light to merge.

rsolomakhin (approved):

index.html
```diff
@@ -561,14 +565,36 @@
...
+                allowing adding new payment instruments for the <a data-cite=
+                "!HTML#current-settings-object">current settings object</a>'s
+                origin is acceptable. If it is, set <var>permission</var> to
+                "granted", and "denied" otherwise.
+                </li>
+                <li>Resolve <var>p</var> with <var>permission</var>.
+                </li>
+              </ol>
+            </li>
+            <li>Return <var>p</var>
```
adrianhopebailie, line 595: Shouldn't we return the promise before we resolve it?

adrianhopebailie (changes requested):

index.html
romandev: @adrianhopebailie Thank you for review.

As far as I know, we are already using the pattern in many places in this spec or other specs. [1][2]
Resolving the promise will happen in parallel due to the following sentence:
"Run the following steps in parallel:"

So, I think it makes sense. WDYT?
(If I'm missing something, please correct me! :) )

[1] https://w3c.github.io/payment-handler/#has-method
[2] https://w3c.github.io/ServiceWorker/#cache-match

adrianhopebailie: You are right, happy for you to merge as is. 

I find the pattern used in the Payment Request spec easier to follow. If you agree then perhaps we should change across the document in a new PR?

1. Let `p` be a new Promise
2. Return `p` and perform the remaining steps in parallel.

Example in https://www.w3.org/TR/payment-request/#show-method at step 9.

adrianhopebailie (approved):

index.html
romandev: I'll do it in follow-up PR!
Could you merge this PR? I don't have a permission to merge.
Thank you!

PR #185, opened 2017-06-23
Fix aligns of examples for image.src.

romandev:
It causes |tidy| bugs. For example, empty lines are added continuously whenever
running |tidy| command with our configuration. We are using |wrap: 80| option
in tidyconfig.txt but it happens because the length of title of example is over
80 bytes. This PR is fixing the problem by reducing the lines.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1245,56 +1245,52 @@
           different execution contexts.
         </p>
         <pre class="example html" title=
-        "Resolving the relative URL of image.src in execution context of window object.">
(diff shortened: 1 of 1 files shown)


romandev:
@ianbjacobs PTAL

You already tried to remove the empty lines but it still makes more empty lines.
whenever running tidy command. It looks a tidy bug. :(

ianbjacobs (approved):

PR #186, opened 2017-06-23
Change the patterns of returning promise to more readable patterns.

romandev:
This PR is initiated from #180. We should change the patterns of returning
promise to more readable patterns as follows:
  BEFORE
    Let p be a new promise.
    Run the following steps in parallel:
      Do something
      Resolve p
    return p

  AFTER
    Let p be a new promise.
    Return p and perform the remaining steps in parallel:
    Do something
    Resolve p

FYI, the patterns are already widely used in PaymentRequest API spec.

### DIFF
--- a/index.html
+++ b/index.html
@@ -590,24 +590,21 @@
           <ol>
             <li>Let <var>p</var> be a new promise.
             </li>
-            <li>Run the following steps in parallel:
-              <ol>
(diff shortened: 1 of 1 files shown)


romandev:
@adrianhopebailie, @ianbjacobs  PTAL

adrianhopebailie (approved):
LGTM

PR #188, opened 2017-07-21
Remove wallets, clean up issues

adrianhopebailie:
Per WG resolution on 20 July, removing wallets
Clean up issue markers based on latest issue status


ianbjacobs:
Hi Adrian,

Thanks for writing this up. Let's discuss at the 24 July payment app task force call.

I think we also need further discussion on "how to do N wallets" in the absence of
the Wallet feature, and furthermore, whether that even makes sense if, for example,
the (display) feature is not going to be implemented.

index.html
```diff
@@ -1,6 +1,8 @@
 <!DOCTYPE html>
 <html>
   <head>
+    <meta name="generator" content=
```
marcoscaceres: @adrianhopebailie, make sure when you run tidy, you run it like this:

```
tidy -config tidyconfig.txt -o index.html index.html
```

marcoscaceres: (it didn't seem to pick up the config file)

index.html
```diff
@@ -480,10 +477,7 @@
             A Web payment app is associated with an origin. Payment handlers
             respond to PaymentRequestEvents. PaymentManagers manage the
             definition, display, and user selection of PaymentInstruments. A
-            PaymentInstrument supports one or more payment methods. A
-            PaymentWallet contains one or more PaymentInstruments and may be
-            used to enhance the user experience (e.g., business wallet and
-            personal wallet).
+            PaymentInstrument supports one or more payment methods.
```
marcoscaceres: nit: `<a>PaymentInstrument</a>`

Might as well fix it while we are at here ;)

marcoscaceres: There are also, PaymentManagers, PaymentRequestEvents that should be linked.

marcoscaceres:
I wonder if this get re-tidy'd, if it will show less things being added. It's hard to review right now because the diff thinks lots of things were added.

index.html
marcoscaceres: Actually, just checked... you might need to remove this manually now.

adrianhopebailie:
@marcoscaceres I have run tidy as you suggest but it doesn't change anything. Seems to just be an unfortunate diff that has a lot of false positives

marcoscaceres:
@adrianhopebailie thanks for checking. :shipit:!

marcoscaceres (approved):

ianbjacobs:
Hi all,

I plan to review at the payment apps task force call on 25 July, then if all is well, merge.

Ian

PR #193, opened 2017-07-25
Add missing </section> after #188 PR.

romandev:
Due to the missing tag, indexes are broken.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1053,218 +1053,216 @@
      </pre>
         </section>
       </section>
-      <section id="instrument-display-ordering">
+    </section>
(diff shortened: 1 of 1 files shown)


romandev:
@ianbjacobs PTAL

ianbjacobs:
Hi @romandev,

I suspect that your change is a single line of HTML, but the diff is huge. Can you help me understand why that is?

Ian

adrianhopebailie (approved):
This is why the diff on PR 188 was so ugly. Weird that Tidy didn't catch this 🤦‍♂️

Issue #190, opened 2017-07-24
respondWith behavior not specified


mkruisselbrink:
Unless I'm missing something, I don't see anything in the spec that (formally or even informally) describes what the respondWith method is supposed to do? All I see is https://w3c.github.io/payment-handler/#respondwith()-method, but that doesn't actually have any described behavior, just what it is used for.

mkruisselbrink:
https://w3c.github.io/payment-handler/#details-attribute seems to have some further text. Although it seems to have a MUST requirement to run an algorithm if the promise is rejected, but the algorithm is left completely up to implementers? That doesn't really make sense. It can't be a MUST requirement if there isn't any actually required behavior.

Ah, and I guess the next paragraph is describing the behavior if the promise is resolved. Maybe the respondWith description could be extended to actually link to the spec text describing its behavior? I'm also not sure how things like failures during the structured clone (or the supposed json-serializable requirements) are supposed to be dealt with by implementations. More formal algorithmic descriptions of the behavior of respondWith would be really helpful.

ianbjacobs:
Closing this after merge of pull request #194.

Ian

Issue #191, opened 2017-07-24
"PaymentRequestEvent dictionary" makes no sense


mkruisselbrink:
In https://w3c.github.io/payment-handler/#details-attribute the text talks about a "PaymentRequestEvent dictionary". But PaymentRequestEvent is an interface... Is that just a typo or was something else entirely intended?

ianbjacobs:
Closed with merge of pull request #194.

PR #194, opened 2017-07-25
Define PaymentRequestEvent.respondWith() behavior.

romandev:
The current spec says outline of the method's behavior but it is not specific.
This patch makes the behavior of the method clearer. (#190, #191)

### DIFF
--- a/index.html
+++ b/index.html
@@ -1095,7 +1095,7 @@
           readonly attribute FrozenArray&lt;PaymentDetailsModifier&gt; modifiers;
           readonly attribute DOMString instrumentKey;
           Promise&lt;WindowClient?&gt; openWindow(USVString url);
-          void respondWith(Promise&lt;PaymentHandlerResponse&gt;handlerResponse);
+          void respondWith(Promise&lt;PaymentHandlerResponse&gt;handlerResponsePromise);
         };
       </pre>
         <section>
@@ -1199,12 +1199,15 @@
         </section>
         <section>
           <h2>
-            <dfn>respondWith</dfn>() method
+            <dfn data-lt=
+            "respondWith(handlerResponsePromise)">respondWith()</dfn> method
           </h2>
           <p>
             This method is used by the payment handler to provide a
             <a>PaymentHandlerResponse</a> when the payment successfully
-            completes.
+            completes. When called, it runs the <a>Respond to PaymentRequest
+            Algorithm</a> with <var>event</var> and
+            <var>handlerResponsePromise</var> as arguments.
           </p>
         </section>
         <p class="issue" title="Share user data with payment app?" data-number=
@@ -1391,6 +1394,17 @@
               This value is a result of <a data-cite=
               "!appmanifest#fetching-image-objects">fetching image object</a>
               or a fallback image provided by the user agent.
+            </td>
+          </tr>
+          <tr>
+            <td>
+              <dfn>[[\respondWithCalled]]</dfn>
+            </td>
+            <td>
+              false
+            </td>
+            <td>
+              YAHO
             </td>
           </tr>
         </table>
@@ -1709,9 +1723,9 @@ window.addEventListener("message", function(e) {
             The user agent receives a successful response from the payment
             handler through resolution of the Promise provided to the
             <a data-lt="PaymentRequestEvent.respondWith">respondWith()</a>
-            function of the corresponding <a>PaymentRequestEvent</a>
-            dictionary. The application is expected to resolve the Promise with
-            a <a>PaymentHandlerResponse</a> instance containing the payment
+            function of the corresponding <a>PaymentRequestEvent</a> interface.
+            The application is expected to resolve the Promise with a
+            <a>PaymentHandlerResponse</a> instance containing the payment
             response. In case of user cancellation or error, the application
             may signal failure by rejecting the Promise.
           </p>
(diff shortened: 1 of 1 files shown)


romandev:

ianbjacobs:
Hi @romandev,

This looks like very good work. I feel I am not the best person to review this and so would like to hear from @rsolomakhin and anyone else who has more experience on these paradigms. Thank you!

Ian

index.html
```diff
@@ -1918,40 +1957,110 @@
       </section>
       <section>
         <h2>
-          Extention to User Accepts the Payment Request Algorithm
```
rsolomakhin, line 1736: Should the spec say somewhere that "Respond to PaymentRequest Algorithm" should be running as part of the "User Accepts the Payment Request Algorithm" in https://w3c.github.io/payment-handler/?

rsolomakhin: Found it below.

rsolomakhin (approved):
LGTM

ianbjacobs:
@mkruisselbrink, as this relates to the issue you raised (#190) it would be great to hear whether you feel this change addresses your issue (and thus, if merged, can we close #190?).

Thank you,

Ian

index.html
```diff
@@ -1918,40 +1957,110 @@
...
-        <ol>
-          <li>Set <var>handlerResponse</var> to the
-          <a>PaymentHandlerResponse</a> instance used to resolve the
-          <a>PaymentRequestEvent</a>.<a data-lt=
-          "PaymentRequestEvent.respondWith">respondWith()</a> Promise.
+        <ol class="algorithm">
+          <li>If <var>event</var>'s <a>dispatch flag</a> is unset, then
+          <a>throw</a> an "<a>InvalidStateError</a>" <a>DOMException</a> and
+          abort these steps.
           </li>
```
mkruisselbrink, line 1761: Should there be a check here, or somewhere else, to make sure respondWith is only called once (i.e. similar to FetchEvent's "respond-with entered flag")?

index.html
```diff
@@ -1918,40 +1957,110 @@
...
-          Extention to User Accepts the Payment Request Algorithm
+          <dfn>Respond to PaymentRequest Algorithm</dfn>
         </h2>
         <p>
-          If the Promise is successfully resolved, the user agent MUST run the
-          <a>user accepts the payment request algorithm</a> as defined in
-          [[!payment-request]], replacing steps 6 and 7 with these steps or
-          their equivalent:
+          When this algorithm with <var>event</var> and
+          <var>handlerResponsePromise</var> parameters are invoked, the user
```
mkruisselbrink: This should at least be "When this algorithm [...] is invoked", and maybe even re-order it to "When this algorithm is invoked with `event` and `handlerResponsePromise` parameters, "

index.html
```diff
@@ -1918,40 +1957,110 @@
...
+                  replacing steps 7 and 8 with these steps or their equivalent.
+                    <ol>
+                      <li>Create a <a>structured clone</a> of
+                      <var>handlerResponse</var>.<a data-lt=
+                      "PaymentHandlerResponse.methodName">methodName</a> and
+                      assign it to associated PaymentRequest's <a data-lt=
+                      "PaymentResponse"><var>response</var></a>.<a data-cite=
+                      "!payment-request#dom-paymentresponse-methodname">methodName</a>.
+                      </li>
+                      <li>Create a <a>structured clone</a> of
```
mkruisselbrink: "structured clone" does not exist anymore in the latest HTML spec (because the way it was defined was problematic with which realms things are done in). Instead you'll need to call [StructuredSerialize](http://w3c.github.io/html/infrastructure.html#section-structuredserialize) in the context of the service worker (i.e. before calling the "user accepts the payment request algorithm"), and then [StructuredDeserialize](http://w3c.github.io/html/infrastructure.html#section-structureddeserialize) here, to create javascript objects in the correct realm. One reason for that is because serialization can have observable side effects in the source context, so it needs to be well specified when serialization happens.

index.html
```diff
@@ -1918,40 +1957,110 @@
...
-          <li>If <var>handlerResponse</var>.<var>details</var> is not present,
-          run the <a>payment app failure algorithm</a> and terminate these
-          steps.
+          <li>Increment the <var>event</var>'s <a>pending promises count</a> by
+          one.
           </li>
-          <li>Create a <a>structured clone</a> of
-          <var>handlerResponse</var>.<var>details</var> and assign it to <var>
-            response</var>.<var>details</var>.
+          <li>Perform the remaining steps in parallel:
```
mkruisselbrink: I don't think you need "perform the remaining steps in parallel", or the "wait until promis settles" below. Instead you can just directly have 
1. Upon rejection of handlerResponsePromise:
   1. Do stuff
1. Upon fulfillment of handlerResponsePromise:
   1. Do stuff
1. Upon fulfillment or rejection of handlerResponsePromise queue a microtask to perform the following steps:
   1. Do stuf

mkruisselbrink (commented):
This definitely looks a lot better than what was there before. There still is the weirdness of saying that the user agent MUST run the "payment app failure" algorithm, without any requirement as to what that algorithm does. It's not really a MUST requirement if there aren't any actual requirements.

marcoscaceres (dismissed):
Needs tests

romandev:
I'll update this PR in this week. (including tests if possible)

romandev (commented):
Rebased and updated.

ianbjacobs:
@mkruisselbrink  and @romandev are you satisfied with the state of this pull request? Thanks!

Ian

romandev:
@ianbjacobs Sorry for delayed reply.
I'm attending BlinkOn8 in Tokyo without note pc.
Anyway I addressed all comments except payment app failure algorithm.
The algorithm is unclear now but we can improve it in follow-up PR.
Thanks.

PR #197, opened 2017-08-02
Add "Validate Methods Algorithm"

rsolomakhin:
Validate the payment method identifiers that a service worker can use
depending on their validity and the corresponding payment method
manifests.

### DIFF
--- a/index.html
+++ b/index.html
@@ -62,6 +62,20 @@
             wgPatentURI:  "https://www.w3.org/2004/01/pp-impl/83744/status",
             issueBase:    "https://github.com/w3c/payment-handler/issues/",
             githubAPI:    "https://api.github.com/repos/w3c/payment-handler",
+            localBiblio: {
+                "payment-method-manifest": {
+                    "href": "https://w3c.github.io/payment-method-manifest/",
+                    "title": "Payment Method Manifest",
+                    "publisher": "W3C",
+                    "status": "LD",
+                    "authors": [
+                        "Dapeng Liu",
+                        "Domenic Denicola",
+                        "Zach Koch"
+                    ],
+                    "rawDate": "2017-05-30"
+                }
+            },
         };
     </script>
     <style>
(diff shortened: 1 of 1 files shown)


rsolomakhin:
I've created this pull request to solicit comments on the outlined approach for using Payment Method Identifier and Payment Method Manifest specs in here. Please let me know what you think. I will be getting some implementation experience in the coming weeks and will report back.

ianbjacobs:
Thanks @rsolomakhin, this proposal relates to my security question:
 https://github.com/w3c/payment-method-manifest/issues/11

Ian

rsolomakhin:
@ianbjacobs wrote:

> The browser implements Payment Request API but not Payment Method Manifest

I've addressed this issue in this pull request by checking whether [Payment Method Manifest](https://w3c.github.io/payment-method-manifest/) spec is implemented. In this case, payment handlers can use all [standardized payment method identifiers](https://www.w3.org/TR/payment-method-id/#dfn-standardized-payment-method-identifier), but as for [URL-based payment method identifiers](https://www.w3.org/TR/payment-method-id/#dfn-url-based-payment-method-identifier), the [origin](https://tools.ietf.org/html/rfc6454) of the identifier must be equal to the [origin](https://tools.ietf.org/html/rfc6454) of the [service worker registration](https://www.w3.org/TR/service-workers-1/#service-worker-registration-concept) [scope URL](https://www.w3.org/TR/service-workers-1/#dfn-scope-url).

rsolomakhin:
@ianbjacobs wrote:

> The browser implements Payment Method Manifest, but there is a failure of some
sort (related to fetching, parsing, or ingesting the manifest file).

I've explicitly added a check for such failure.

marcoscaceres:
Please be sure to provide WPT tests along with all PRs.

rsolomakhin:
Looks like PMM spec already has this algorithm. What we need to say in the PH spec is that
when the browser matches payment method manifest origin against the payment app origin, it uses the payment handler scope URL origin.

Issue #198, opened 2017-08-03
`requestPermission` is static in Web IDL but example is not


dlongley:
While working on implementing a polyfill for the API I noticed that the [spec](https://www.w3.org/TR/payment-handler/#paymentmanager-interface) indicates that the `requestPermission` call on the `PaymentManager` interface is `static`. However, the [example](https://www.w3.org/TR/payment-handler/#register-example) calls it on the service worker registration's `paymentManager` instance (rather than directly on the class).

Should the example be fixed or the Web IDL?

adrianhopebailie:
Resolved by https://github.com/w3c/payment-handler/pull/199

PR #199, opened 2017-08-04
Fix a example for PaymentManager.requestPermission().

romandev:
Although PaymentManager.requestPermission is defined as static method, its
example is using the service worker registration's paymentManager instance.
It's better to use the static method in the example if possible.

This PR is fixing #198 issue.

### DIFF
--- a/index.html
+++ b/index.html
@@ -525,7 +525,7 @@
           <dfn>PaymentManager</dfn> interface
         </h2>
         <pre class="idl">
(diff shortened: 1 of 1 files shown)


romandev:
@dlongley, @ianbjacobs PTAL

dlongley (approved):
LGTM

ianbjacobs (approved):

Issue #201, opened 2017-08-10
Change openWindow algorithm 14.1 to 'Resolve promise with null.'


gogerald:
At this point the window is opened, so the openWindow is succeed, but because it opens a crossing origin window, so there is no WindowClient could return.

Any thoughts?

dlongley:
It should not be a cross origin window. It should be opening a window to the same origin that the service worker is running in.

If the window opens an origin that isn't the same as the service worker's then a `"SecurityError" DOMException` is thrown ([see step 14 of the open window algorithm](https://w3c.github.io/payment-handler/#dfn-open-window-algorithm)).

gogerald:
But that doesn't make too much sense since the window is already opened at that point. I guess this is why we return null in Clients.openWindow for service worker at the same situation. Might be good to keep them consistent to make it easier for implementation as well.

I created a pull request #202 . Feel free to comment on it.

gogerald:
@rsolomakhin @ianbjacobs

romandev:
Seems reasonable to me. I thought it's a little bit strange.

Even if openWindow() takes a normal origin url, it may eventually end up with a cross origin url due to server side redirection. In this case, openWindow() will reject with SecurityError as per current spec but Browser Tab(or Browser Window) is already shown to user. So, although the method is rejected, it will show about:blank or a cross origin page. (@gogerald's point)

Another case is a bit more weird. Please see the following code.
```
e.openWindow(normal_origin_url)
    .then(windowClient => {
        // openWindow is succeeded but...
        windowClient.navigate(cross_origin_url);
    });
```

Even if openWindow() is succeeded, the opened window may display a cross origin url due to windowClient.navigate() or location.href.

So, it's too difficult to block a cross origin url during openWindow especially navigation.
I think a ServiceWorker's way (resolving null) is reasonable and consistent.

ianbjacobs:
Hi @gogerald and @romandev, is this issue closable followinfg the merge of pull request #202?

Ian

romandev:
Sure, this is already fixed by @gogerald. I'll close this issue :)

PR #202, opened 2017-08-10
Resolve with null instead of reject with 'SecurityError' if PaymentRequestEvent.openWindow is opened in a different origin

gogerald:
In the openWindow algorithm 14.1, the window is opened at that point, looks make no sense to reject with "SecurityError". 

This is also how Clients.openWindow works for service work.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1707,7 +1707,7 @@
           </li>
           <li>If <var>url</var>'s origin is not the same as the <a>service
           worker</a>'s origin associated with the payment handler, return a <a>
-            Promise</a> rejected with a <a>SecurityError</a>.
+            Promise</a> resolved with null.
           </li>
           <li>If this algorithm is not <a data-cite=
           "!HTML#triggered-by-user-activation">triggered by user
--- a/index.html
+++ b/index.html
@@ -1262,7 +1262,7 @@
           readonly attribute object total;
           readonly attribute FrozenArray&lt;PaymentDetailsModifier&gt; modifiers;
           readonly attribute DOMString instrumentKey;
-          Promise&lt;WindowClient&gt; openWindow(USVString url);
+          Promise&lt;WindowClient?&gt; openWindow(USVString url);
           void respondWith(Promise&lt;PaymentHandlerResponse&gt;handlerResponse);
         };
       </pre>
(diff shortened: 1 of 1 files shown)


gogerald:
@rsolomakhin @ianbjacobs @romandev what you think?

romandev:
@gogerald, I left my opinion in #201.
This patch looks good to me but you should fix IDL as well I think. (Should be nullable type)

romandev:
Also, we can remove step 4.

gogerald:
Thanks. Why remove step 4 'If the url parsing throws an exception, return a Promise rejected with that exception.'? It looks make sense to keep it.

romandev:
Ah, sorry for confusing. I meant step 6:
"If url's origin is not the same as the service worker's origin associated with the payment handler, return a Promise rejected with a SecurityError."

IMHO, for consistency, we should resolve null instead of rejecting with SecurityError.

gogerald:
It makes sense to me, I removed that check,

rsolomakhin (approved):
Looks good to me. Please add a link to the relevant Service Worker algorithm in the commit description:

https://w3c.github.io/ServiceWorker/#clients-openwindow

index.html
```diff
@@ -1705,10 +1705,6 @@
           <li>If <var>url</var> is <code>about:blank</code>, return a
           <a>Promise</a> rejected with a <a>TypeError</a>.
           </li>
-          <li>If <var>url</var>'s origin is not the same as the <a>service
-          worker</a>'s origin associated with the payment handler, return a <a>
-            Promise</a> rejected with a <a>SecurityError</a>.
-          </li>
```
rsolomakhin: I think we should keep this check, as well, but we can return `null` for consistency. This check is to make sure that service worker from`https://bobpay.xyz` does not try to show a page from `https://alicepay.xyz`. This can be checked synchronously, so there's no need to wait until the page opens.

gogerald: https://bobpay.xyz can do window.location.href = 'https://alicepay.xyz' immediately, so with this check we still can not prevent it happen. 

My understanding is that we no need to do any origin related check (if it is user triggered) for openWindow.

The returned "The WindowClient interface of the ServiceWorker API represents the scope of a service worker client that is a document in a browser context, controlled by an active worker."

If the window is opened, redirected or navigated out of the origin, then the service worker has no control of it. We can not find that window through Clients.matchAll.

Make sense?

rsolomakhin:
Great, thank you!

romandev (approved):
lgtm2

PR #204, opened 2017-08-21
Simplify Payment Handler Registration example.

### LINKED ISSUE #200: Registration example accesses `registration` before it is defined
In the [registration example](https://www.w3.org/TR/payment-handler/#register-example) the service worker `registration` variable is accessed before it is defined. I recommend moving the registration retrieval and `paymentManager` feature check to the top of the method:

```js
button.addEventListener("click", async() => {
  const { registration } =
    await navigator.serviceWorker.register('/sw.js');
  if (!registration.paymentManager) {
    return; // not supported, so bail out.
  }

  const permission =
    await navigator.permissions.query({ name: "paymenthandler" });
  switch (permission) {
    case "denied":
      return;
    case "prompt":
      const result = await registration.paymentManager.requestPermission();
      if (result !== "granted") {
        return;
      }
      break;
  }

  // Excellent, we got it! Let's now set up the user's cards.
  await addInstruments(registration);
}, { once: true });
```

dlongley:
It is unnecessary to call `navigator.permissions.query`. Instead
only the `PaymentManager.requestPermission` call is made. Also,
the feature check for `PaymentManager` is made earlier.

Fixes #200.

### DIFF
--- a/index.html
+++ b/index.html
@@ -995,24 +995,17 @@
           </p>
           <pre class="example js" title="Payment Handler Registration">
         button.addEventListener("click", async() =&gt; {
-          const permission =
-            await navigator.permissions.query({ name: "paymenthandler" });
-          switch (permission) {
-            case "denied":
-              return;
-            case "prompt":
-              const result = await PaymentManager.requestPermission();
-              if (result !== "granted") {
-                return;
-              }
-              break;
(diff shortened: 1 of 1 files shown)


romandev (approved):
lgtm

PR #207, opened 2017-08-28
Add AbortPaymentEvent.

rsolomakhin:
***
<a href="https://pr-preview.s3.amazonaws.com/rsolomakhin/payment-handler/pull/207.html" title="Last updated on Nov 28, 2017, 9:00 AM GMT">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-handler/c8d8a6b...rsolomakhin:4b83039.html" title="Last updated on Nov 28, 2017, 9:00 AM GMT">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -46,6 +46,7 @@
                     w3cid:      2175
                 },
                 {   name:       "Rouslan Solomakhin",
+                    url:        "https://github.com/rsolomakhin",
                     company:    "Google",
                     companyURL: "https://www.google.com/",
                     w3cid:      83784
@@ -199,7 +200,8 @@
           <dfn data-cite=
           "!payment-request#paymentmethoddata-dictionary">PaymentMethodData</dfn>,
           <dfn data-cite="!payment-request#id-attribute">ID</dfn>,
-          <dfn data-cite="!payment-request#show-method">show()</dfn>, and
+          <dfn data-cite="!payment-request#show-method">show()</dfn>,
+          <dfn data-cite="!payment-request#abort-method">abort()</dfn>, and
           <dfn data-cite=
           "!payment-request#user-accepts-the-payment-request-algorithm">user
           accepts the payment request algorithm</dfn> are defined by the
@@ -398,6 +400,14 @@
         <a>payee</a>. If interaction with the user is necessary, the <a>payment
         handler</a> can open a window for that purpose.
         </li>
+        <li>The <a>payee</a> may choose to limit the time window of the
+        availability of their offer by calling [[payment-request]] method
+        <a>abort()</a>, which is forwarded to the <a>payment handler</a> as <a>
+          AbortPaymentEvent</a>. The <a>payment handler</a> can respond with a
+          boolean that indicates whether the payment was successfully aborted.
+          No expectation of transaction rollback or refunds is placed on the
+          <a>payment handler</a>.
+        </li>
         <li>The user agent receives a response asynchronously once the payment
         handler has finished handling the request. The response becomes the
         PaymentResponse (of [[!payment-request]]).
@@ -517,7 +527,11 @@
         partial interface ServiceWorkerRegistration {
           readonly attribute PaymentManager paymentManager;
         };
-      </pre>
+        </pre>
+        <p>
+          The <a>paymentManager</a> attribute exposes payment handler
+          functionality in the service worker.
+        </p>
       </section>
       <section data-dfn-for="PaymentManager" data-link-for="PaymentManager">
         <h2>
@@ -1208,15 +1222,6 @@
         <a>PaymentHandlerResponse</a> to create a PaymentReponse for
         [[!payment-request]].
       </p>
-      <p class="issue" title=
-      "Support for Abort() being delegated to Payment Handler" data-number=
-      "117">
-        Payment Request API supports delegation of responsibility to manage an
-        abort to a payment app. There is a proposal to add a
-        paymentRequestAborted event to the Payment Handler interface. The event
-        will have a respondWith method that takes a boolean parameter
-        indicating if the paymentRequest has been successfully aborted.
-      </p>
       <section data-dfn-for="ServiceWorkerGlobalScope" data-link-for=
       "ServiceWorkerGlobalScope">
         <h2>
@@ -1357,7 +1362,7 @@
         </section>
         <section>
           <h2>
-            <dfn>openWindow</dfn>() method
+            <dfn>openWindow()</dfn> method
           </h2>
           <p>
             This method is used by the payment handler to show a window to the
@@ -1366,7 +1371,7 @@
         </section>
         <section>
           <h2>
-            <dfn>respondWith</dfn>() method
+            <dfn>respondWith()</dfn> method
           </h2>
           <p>
             This method is used by the payment handler to provide a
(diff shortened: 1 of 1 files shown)


rsolomakhin:
Pease don't merge until I've reported on implementation experience.

marcoscaceres:
Made a "do not merge" label... to celebrate :)

romandev:
If CanMakePayment event is risky, AbortPayment event might also be risky?
Because it can be triggered from merchant's abort() call without user interaction.
The event doesn't take any information but it's not important.
Might take some informations such as urls by using clients.matchAll with includeUncontrolled option.

rsolomakhin:
The user agent fires the `AbortPayment` event only in the currently invoked payment app. From user perspective, that's after the user has selected a payment app and clicked [PAY], while the user is waiting for the payment app to complete. From implementation perspective, that's after `PaymentRequest` event has been fired, but before `PaymentRequest.updateWith()` promise has been resolved. At this point, the user has given explicit consent for the merchant website's information to be passed into the payment app to facilitate the payment. Hence, the risk of `AbortPayment` event is low.

index.html
```diff
@@ -517,7 +527,11 @@ <h2>
         partial interface ServiceWorkerRegistration {
           readonly attribute PaymentManager paymentManager;
         };
-      </pre>
+        </pre>
+        <p>
+          The <a>paymentManager</a> attribute exposes payment handler
```
marcoscaceres, line 532: Shouldn't this be a `dfn`?

index.html
```diff
@@ -1641,6 +1646,137 @@ <h2>
         </ol>
       </section>
     </section>
+    <section id="abort">
+      <h2>
+        Aborting an invocation
+      </h2>
+      <p>
```
marcoscaceres, line 1653: This paragraph is redundant. This should be in an algorithm somewhere (probably is, just haven't gotten to it).

index.html
```diff
@@ -1641,6 +1646,137 @@ <h2>
...
+        If the <a>payee</a> uses [[!payment-request]] method <a>abort()</a>,
+        the <a>user agent</a> fires <a>AbortPaymentEvent</a> in the <a>payment
+        handler</a> that is currently handling <a>PaymentRequestEvent</a>.
+        There is no expectation that the <a>payment handler</a> would roll back
+        a transaction or issue a refund. (This is out of scope.) The <a>payment
```
marcoscaceres, line 1658: Note, this whole talk of "refunds" and other "money talk" is not ideal. We should only talk about how we manipulate things in the API - but not go into this kind of detail. We can describe "here is how you do a refund" in tutorial material.

index.html
```diff
@@ -1641,6 +1646,137 @@ <h2>
...
+        handler</a> should respond with a boolean that indicates whether the
+        abort was successful.
+      </p>
+      <p>
+        Implementations may impose a timeout for developers to respond to the
+        <a>AbortPaymentEvent</a>. If the timeout expires, then the
+        implementation will behave as if <a data-lt=
```
marcoscaceres, line 1665: Nit: there is no need to use `data-lt` :) Just add `data-link-for="AbortPaymentEvent"` in a parent element (either the paragraph or the parent `<section>`).

index.html
```diff
@@ -1641,6 +1646,137 @@ <h2>
...
+        partial interface ServiceWorkerGlobalScope {
+          attribute EventHandler onabortpayment;
+        };
+        </pre>
+        <section>
+          <h2>
+            <dfn>onabortpayment</dfn> attribute
+          </h2>
+          <p>
+            The <a>onabortpayment</a> attribute is an <a>event handler</a>
```
marcoscaceres, line 1688: Nit: `<a>event handler</a>` should be `<a>event handler attribute</a>` and linked to HTML correctly.

index.html
```diff
@@ -1641,6 +1646,137 @@ <h2>
...
+            whose corresponding <a>event handler event type</a> is
+            abortpayment.
```
marcoscaceres, line 1690: `abortpayment` should be in quotes

index.html
```diff
@@ -1641,6 +1646,137 @@ <h2>
...
+            was able to abort the payment.
+          </p>
+        </section>
+      </section>
+      <section>
+        <h2>
+          Handling an AbortPaymentEvent
+        </h2>
+        <p>
+          Upon <a data-cite=
```
marcoscaceres, line 1724: Nit: we should try to be consistent with how we reference PaymentRequest.show/abort/etc. See the fix I made above, which gives you the nice linking both the PaymentRequest interface and the method itself..

index.html
```diff
@@ -1641,6 +1646,137 @@ <h2>
...
+          "!payment-request#abort-method">PaymentRequest.abort()</a>, the
+          <a>user agent</a> MUST run the following steps:
+        </p>
+        <ol>
+          <li>Let <var>registration</var> be the
+          <a>ServiceWorkerRegistration</a> corresponding to the
+          <a>PaymentRequestEvent</a> that is being processed.
```
marcoscaceres, line 1731: Q: How are these linked together? Is there some more explicit means to get at the registration from the event?

index.html
```diff
@@ -1641,6 +1646,137 @@ <h2>
...
+          </li>
+          <li>If <var>registration</var> is not found, reject the
```
marcoscaceres, line 1733: "not found" makes is sounds like some kind of lookup took place. Maybe we want:

 1. Let registration be (... steps to get the registration....)
 1. If registration is null, then...

index.html
```diff
@@ -1641,6 +1646,137 @@ <h2>
...
+          <a>Promise</a> that was created by <a data-cite=
```
marcoscaceres, line 1734: So, I don't think we can do this... we need to hook back into Payment Request here - but we can't reach across threads and reject the promise?

index.html
```diff
@@ -1641,6 +1646,137 @@ <h2>
...
+          "!payment-request#abort-method">PaymentRequest.abort()</a> with false
+          and terminate these steps.
+          </li>
+          <li>Invoke the <a>handle functional event</a> algorithm with a
+          <a>ServiceWorkerRegistration</a> of <var>registration</var> and <var>
+            callbackSteps</var> set to the following steps:
```
marcoscaceres, line 1740: Maybe we want some kinda of task here? This looks like a task.

index.html
```diff
@@ -1641,6 +1646,137 @@ <h2>
...
+            <ol>
```
marcoscaceres, line 1741: Yeah, we shouldn't need to do any of this... Payment Request should do this for us. We just need to tell Payment Request that we have aborted, and it should do all this, no?

index.html
```diff
@@ -1641,6 +1646,137 @@ <h2>
...
+              <li>Set <var>global</var> to the global object that was provided
+              as an argument.
+              </li>
+              <li>Create a <a>trusted event</a>, <var>e</var>, that uses the
+              <a>AbortPaymentEvent</a> interface, which does not bubble, cannot
+              be canceled, and has no default action.
+              </li>
+              <li>Dispatch <var>e</var> to <var>global</var>.
+              </li>
+              <li>Wait for all of the promises in the <a>extend lifetime
```
marcoscaceres, line 1751: This feels out of place to me. Shouldn't we get this for free from .waitUntil() without needing to say anything? Hopefully this whole task can go away.

index.html
```diff
@@ -1641,6 +1646,137 @@ <h2>
...
+              promises</a> of <var>e</var> to resolve.
+              </li>
+              <li>If the <a>payment handler</a> has not provided a boolean
+              response, reject the <a>Promise</a> that was created by
+              <a data-cite=
+              "!payment-request#abort-method">PaymentRequest.abort()</a> with a
+              <a>DOMException</a> whose value "<a>OperationError</a>".
```
marcoscaceres, line 1758: You can remove this whole thing. Payment Request doesn't know anything about "OperationError"s and WebIDL will just coerce anything you send into `.respondeWith()` into a boolean.

marcoscaceres: ...anyway, I'm kinda sure this whole thing needs to go away. Fingers crossed!

marcoscaceres (changes requested):
This is a good start. I think we can get Payment Request to do some of the work tho.

index.html
rsolomakhin: There's a <dfn> for paymentManager eight lines above. I'm not super familiar with spec format, but shouldn't there be a single <dfn> tag per term in a spec?

index.html
rsolomakhin: That's a neat trick, thank you!

rsolomakhin:
Let me close this pull request and submit a new one.

PR #209, opened 2017-08-28
Make explicit which origin to use in conjunction with payment method manifest.

rsolomakhin:


index.html
```diff
@@ -2028,6 +2028,17 @@
       </section>
       <section>
         <h2>
+          Supported Origin
+        </h2>
+        <ul>
+          <li>When the user agent matches payment method manifest origin against
+            the payment app origin, it uses the origin of the service worker
+            registration scope URL.
+          </li>
```
ianbjacobs: How about this:

Payment method manifests authorize <a>origins</a> to distribute payment apps for a given payment method. When the user agent is determining whether a payment handler matches the origin listed in a payment method manifest, the user agent uses the scope URL of the payment handler's service worker registration.

rsolomakhin: Done

rsolomakhin:
Please take another loook @ianbjacobs

ianbjacobs (approved):

PR #211, opened 2017-08-29
move dependency section down, as is done in PR API

ianbjacobs:


marcoscaceres (approved):
Eventually, ReSpec will generate this.

PR #214, opened 2017-09-05
Security addition

ianbjacobs:
As discussed during 5 September task force call: it is ok for user agents to not display a (matching) payment app due to security reasons.

(Also some tidying happening here of previous merged pull request about payment instruments from Rouslan.)

Ian

### DIFF
--- a/index.html
+++ b/index.html
@@ -1824,8 +1824,10 @@ window.addEventListener("message", function(e) {
         </h2>
         <ul>
           <li>The user agent is not required to make available payment handlers
-          that pose security issues and should inform the user when a payment
-          handler is unavailable for security reasons.
+          that pose security issues. When a payment handler is unavailable for
+          security reasons, the user agent should provide rationale to the
+          payment handler developers (e.g., through console messages) and may
+          also inform the user to help avoid confusion.
           </li>
         </ul>
         <p class="note">
--- a/index.html
+++ b/index.html
@@ -1472,7 +1472,7 @@
               <li>Set the <a data-lt=
               "PaymentRequestEvent.instrumentKey">instrumentKey</a> attribute
               of <var>e</var> to the <a>instrumentKey</a> of the selected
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1472,7 +1472,8 @@
               <li>Set the <a data-lt=
               "PaymentRequestEvent.instrumentKey">instrumentKey</a> attribute
               of <var>e</var> to the <a>instrumentKey</a> of the selected
-              <a>PaymentInstrument</a>, or an empty string if none was selected.
+              <a>PaymentInstrument</a>, or an empty string if none was
```
marcoscaceres: "the" empty string. https://en.wikipedia.org/wiki/Empty_string

index.html
```diff
@@ -1861,6 +1862,12 @@
         <h2>
           Payment App Authenticity
         </h2>
+        <ul>
```
marcoscaceres, line 1825: Why is this a list?

index.html
```diff
@@ -1861,6 +1862,12 @@
...
+          <li>The user agent is not required to make available payment handlers
```
marcoscaceres, line 1826: This seems rather self evident, tbh. What motivated adding this?

index.html
ianbjacobs: For consistent formatting with the other subsections in this section.

index.html
ianbjacobs: I think there is an expectation that the browser will make available all matching payment apps. This explicitly licenses the browser to exclude payment apps (that would otherwise match) for security reasons.

ianbjacobs:
Hi @rsolomakhin and @adrianhopebailie; I've updated the text based on the 19 Sep payment apps task force call. Review welcome.

https://www.w3.org/2017/09/19-apps-minutes.html

rsolomakhin (approved):

Issue #156, opened 2017-05-12
Establish Editorial process


marcoscaceres:
After FPWD it's imperative we put in place an editorial process that mirrors the Payment Request spec. 

In particular: 

 * Commit message structure (chore:, editorial:, etc.)
 * no review from implementers, no merge. 
 * No merge commits!

Some of these things can be enforced by github.

adrianhopebailie:
Sounds like you are volunteering to be an editor? 😄

ianbjacobs:
I'm ok with that approach, and also with Marcos becoming an editor.

Ian

marcoscaceres:
Yep, I'll definitely help edit.

msporny:
> no review from implementers, no merge.

Editorial changes can still be done w/o implementer review, I assume?

Who is implementing?

Digital Bazaar has committed to do the polyfill implementation after FPWD (sometime during the summer, most likely).

adrianhopebailie:
> Who is implementing?

@rsolomakhin can confirm but I think Google are implementing. I'm hopeful that Mozilla will too since they will soon have 2 editors on the spec and Opera (@tommythorsen) are also editors so I expect they plan to as well.

rsolomakhin:
Google is implementing Payment Handler spec.

marcoscaceres:
We (Moz) are going to investigate the feasibility of implementing over the next few months. There are similarities between this specification and foreign fetch, which is challenging for us (Mozilla) given some limitations in our current implementation of service workers. 

However, we remain optimistic that we will be able to implement and remain supportive of this work - and we intend to continue to work with the fine folks in this WG towards creating a great solution for the web to enable web apps to handle requests for payment.

marcoscaceres:
Better: https://whatwg.org/working-mode

ianbjacobs:
I am going to close this issue as we have a functioning editorial process; can reopen if needed.

PR #224, opened 2017-11-27
Only allows instrument.set() in active service worker

### LINKED ISSUE #223: SW registration code unclear/incorrect
The sample service worker registration code included in the spec is potentially misleading in that it gives the impression that a payment handler is ready to handle payments before it actually is.

The registration example code at https://w3c.github.io/payment-handler/#register-example registers a service worker in a typical way:

```js
  const { registration } = await navigator.serviceWorker.register('/sw.js');
  await addInstruments(registration);
```

My understanding is that after these two lines have been executed, the service worker defined in `sw.js` is not necessarily active: the service worker controlling the page may be a previous service worker (one not able to handle payment-specific events such as `onpaymentrequest`) and therefore at this point it would *not* be appropriate to tell that user that they can e.g. "Now Pay with Bob Pay."

Whilst the sample code doesn't not explicitly make this claim, I think many readers unfamiliar with the details of the service worker lifecycle would assume that after the service worker is registered and instruments added, the user is ready to go with Bob Pay. Instead, a [more complicated algorithm](https://github.com/w3c/ServiceWorker/issues/770) must be followed to determine the point at which `sw.js` is active.

/cc @jakearchibald

romandev:
Like other service worker family features, we should only allow
instrument.set() in active service worker.

This fixes #223 issue.

***
<a href="https://pr-preview.s3.amazonaws.com/romandev/payment-handler/pull/224.html" title="Last updated on Oct 2, 2018, 5:08 PM GMT (c0cd0a4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-handler/224/3513adc...romandev:c0cd0a4.html" title="Last updated on Oct 2, 2018, 5:08 PM GMT (c0cd0a4)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -474,6 +474,14 @@
             When called, this method executes the following steps:
           </p>
           <ol>
+            <li>Let <var>registration</var> be the <a>PaymentInstrument</a>'s
+            associated <a>service worker registration</a>.
+            </li>
+            <li>If <var>registration</var> has no <a>active worker</a>, then
+            reject a <a>Promise</a> with an
+            "<code><a>InvalidStateError</a></code>" DOMException and terminate
+            these steps.
+            </li>
             <li>Upon user agent discretion and depending on user consent,
             optionally return a <a>Promise</a> rejected with a
             <a>NotAllowedError</a>.
@@ -716,7 +724,8 @@
         &lt;script&gt;
 
         const instrumentKey = "c8126178-3bba-4d09-8f00-0771bcfd3b11";
-        const registration = await navigator.serviceWorker.register("/register/sw.js");
+        navigator.serviceWorker.register("/register/sw.js");
+        const registration = await navigator.serviceWorker.ready;
         await registration.paymentManager.paymentInstruments.set({
           instrumentKey,
           {
@@ -773,8 +782,8 @@
             return; // not supported, so bail out.
           }
 
-          const registration =
-            await navigator.serviceWorker.register("/sw.js");
+          navigator.serviceWorker.register("/sw.js");
+          const registration = await navigator.serviceWorker.ready;
 
           // Excellent, we got it! Let's now set up the user's cards.
           await addInstruments(registration);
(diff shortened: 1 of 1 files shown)


marcoscaceres (commented):
The proposed text seems fine, but does this also apply to things like IndexedDB? Is there a pointer to an issue describing why we are adding this restriction? Just wondering what the rationale is.

ianbjacobs:
@romandev, we discussed this briefly at the editor call today and would like to request additional explanation of why you believe this change address issue #223. Thanks!

romandev:
Thank you for reviews and I'm sorry for delayed reply.
I was on long vacations last week.

This pull request is initiated from #223. The problem would not have happened before (https://github.com/w3c/payment-handler/commit/bfe952bfccf45a696f8b8dd3248cd34053c02ac2).
Because there was a spec text that it should wait for waiting/installing worker to become its active worker. We guessed the spec text was removed mistakenly. So, I'd like to restore the spec text from the large change.

One solution is `navigator.serviceWorker.ready` but it's not perfect.
For examples, there are the following issues:
  - What happens when calling register()/set() in redaundant service worker?
  - What happens in case of multiple registrations?

I think this patch can resolve them.

romandev:
> The proposed text seems fine, but does this also apply to things like IndexedDB?

If my understanding is correct, IndexedDB storage is not associated with a service worker.
We might compare it with CacheStorage, BackgroundSync, and Push rather than IndexedDB.

The following link will be helpful to us:
  - https://github.com/WICG/BackgroundSync/issues/101

marcoscaceres:
> If my understanding is correct, IndexedDB storage is not associated with a service worker.

Ah, got it... it's "exposed" in a service worker, but not associated with it. 

Right - I'm my mind the instruments were not bound to the service worker... that's again the discussion around storage and credentials.

rsolomakhin:
@romandev: Shouldn't all the methods (`.get()`, `.set()`, `.delete()`, `.has()`) work only in an active service worker?

romandev:
@rsolomakhin PTAL

I think other methods(`get()`, `delete()` and so on) no needs the check because instruments storage always empty before calling `set()`.

`get()` will resolve `undefined`.
`delete()` will resolve `false`.
`keys()` will resolve `empty array`.
`has()` will resolve `false`.
`clear()` is unmeaningful because the storage is already empty.

romandev:
FYI, when you wanna merge this or other PRs, according to https://github.com/w3c/payment-handler/issues/156, it's better to use a squash commit. (No merge commits)

I think merge commits make it difficult for us to track our commit history. :)

rsolomakhin (approved):

PR #244, opened 2017-12-18
add a "public computing" section to security/privacy considerations

ianbjacobs:
per:
 https://www.w3.org/2017/12/18-apps-minutes.html#item02

NOTE: 

 * It is assumed that browsers support this, so this is expressed in terms of usage rather than an implementation requirement. However, it is intended to give a heads-up to implementers to be sure to take this into account in any kiosk-type setting.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/payment-handler/pull/244.html" title="Last updated on Dec 18, 2017, 2:52 PM GMT (1544cbd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-handler/244/388c3c3...1544cbd.html" title="Last updated on Dec 18, 2017, 2:52 PM GMT (1544cbd)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1920,6 +1920,17 @@ window.addEventListener("message", function(e) {
           </li>
         </ul>
       </section>
+      <section>
+        <h2>
+          Browsing on Public Computers
+        </h2>
+        <ul>
+          <li>When setting up a public computer or kiosk, it is recommended
+          that browsers be configured not to store sensitive user information,
+          in particular related to payments.
+          </li>
+        </ul>
+      </section>
     </section>
     <section id="dependencies">
       <h2>


ianbjacobs:
Hi @marcoscaceres and @rsolomakhin, does this look ok to you?

marcoscaceres:
This is pretty generic guidance (applies to basically any data), IMO. Seems fairly unnecessary.

ianbjacobs:
@marcoscaceres, I take your point. Maybe I can come up with something better if I have a better understanding of:

 * How people configure browsers generally for public computing settings.
 * Would the addition of payment handler API support change how that is done? If not, then
    there may be nothing to say. But if so, then would a heads-up to implementers to
    take payment handler API into account in "public computing mode" make sense?

Ian

marcoscaceres:
I still think the above is not in scope and distracts from more fundamental issues. For example, we haven’t even come to an agreement about what should happen when a user clears site data. 

Let’s focus on those more fundamental things first, as without solving those we won’t have an API at all. 

Regarding your questions above, I hold that they apply generally to all data held by browsers (e.g., passwords, form autofill, browsing history, etc). It’s not something we are going to solve in this spec. 

The TAG was going to define private browsing, iirc. So it might be something we give to them to document?

ianbjacobs:
The editors discussed this and chose to put information about data clearing hygiene in the developer portal:
 https://github.com/w3c/payment-request-info/wiki/PaymentAppPractice

PR #283, opened 2018-04-09
HTTPS considerations.

rsolomakhin:
Write down the security mitigations in Chrome for other implementers to
be aware.

 * [ ] web platform tests (link)
 * [ ] MDN Docs added

Implementation commitment:

 * [ ] Safari
 * [x] [Chrome](https://crbug.com/828431)
 * [ ] Firefox
 * [ ] Edge

***
<a href="https://pr-preview.s3.amazonaws.com/rsolomakhin/payment-handler/pull/283.html" title="Last updated on Apr 9, 2018, 2:54 PM GMT (30e4b9a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-handler/283/d9ebc35...rsolomakhin:30e4b9a.html" title="Last updated on Apr 9, 2018, 2:54 PM GMT (30e4b9a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1669,6 +1669,15 @@
               </li>
             </ol>
           </li>
+          <li>If the security properties of <var>newContext</var> are
+          unsatisfactory (e.g., invalid certificate), then:
+            <ol>
+              <li>Reject <var>promise</var> with <a>SecurityError</a>..
+              </li>
+              <li>Abort these steps.
+              </li>
+            </ol>
+          </li>
           <li>Let <var>client</var> be the result of running the
             <a data-cite="!SERVICE-WORKERS#create-windowclient-algorithm">create
             window client</a> algorithm with <var>newContext</var> as the
@@ -2077,6 +2086,31 @@ window.addEventListener("message", function(e) {
             was called with <code>true</code>.
           </li>
         </ul>
+      </section>
+      <section>
+        <h2>
+          HTTPS
+        </h2>
+        <ul>
+          <li>The user agent may block mixed content (e.g., non-HTTPS or
+          scripts) on the payment handler page.
+          </li>
+          <li>If the SSL certificate of the payment handler page is not valid
+          (e.g., self-signed), the user agent may cancel the payment.
+          </li>
+          <li>If the web-page is known to be malicious (e.g., a phishing page
+          according to a safe browsing database), the user agent may cancel the
+          payment.
+          </li>
+          <li>If the payment handler page redirects to a non-HTTPS scheme
+          origin, the user agent should cancel the payment.
+          </li>
+        </ul>
+        <p>
+          The user agent should provide rationale to the payment handler
+          developers (e.g., through console messages) and may also inform the
+          user to help avoid confusion whenever these mitigations happen.
+        </p>
       </section>
     </section>
     <section id="display" class="informative">


index.html
```diff
@@ -1669,6 +1669,15 @@ <h2>
               </li>
             </ol>
           </li>
+          <li>If the security properties of <var>newContext</var> are
+          unsatisfactory (e.g., invalid certificate), then:
+            <ol>
+              <li>Reject <var>promise</var> with <a>SecurityError</a>..
```
ianbjacobs, line 1675: editorial: two periods after SecurityError.

index.html
```diff
@@ -2078,6 +2087,31 @@ <h2>
...
+        <h2>
+          HTTPS
+        </h2>
+        <ul>
+          <li>The user agent may block mixed content (e.g., non-HTTPS or
+          scripts) on the payment handler page.
+          </li>
+          <li>If the SSL certificate of the payment handler page is not valid
+          (e.g., self-signed), the user agent may cancel the payment.
+          </li>
```
ianbjacobs, line 2100: I think that is inconsistent with the algorithm definition, which forces abort. I suggest changing this to MUST.

index.html
marcoscaceres, line 1675:
```suggestion
              <li>Reject <var>promise</var> with <a>SecurityError</a>.
```

index.html
```diff
@@ -1669,6 +1669,15 @@ <h2>
               </li>
             </ol>
           </li>
+          <li>If the security properties of <var>newContext</var> are
```
marcoscaceres, line 1672: This seems overly vague. Maybe @domenic can suggest better wording for this? It might be a simple as saying "not a secure context" or something link that.

index.html
```diff
@@ -2078,6 +2087,31 @@ <h2>
           </li>
         </ul>
       </section>
+      <section>
```
marcoscaceres, line 2090: This whole section doesn't feel right in this spec.

marcoscaceres (changes requested):

index.html
domenic: It's unclear from @rsolomakhin's text whether insecure HTTP pages are rejected or not. Invalid certificates will fail a navigation, so that would happen before this step.

If the intent is to disallow insecure HTTP, then the wording would be _newContext_'s [active document](https://html.spec.whatwg.org/multipage/browsers.html#active-document)'s [relevant settings object](https://html.spec.whatwg.org/multipage/webappapis.html#relevant-settings-object) is [contextually secure](https://w3c.github.io/webappsec-secure-contexts/#environment-settings-object-contextually-secure).

rsolomakhin:
Thank you for the review, @domenic , @ianbjacobs , and @marcoscaceres ! My intention here is to block both HTTP (non-secure) pages and HTTPS pages where certificates are self-signed, expired, revoked, and so on. I would encourage user agent implementers to clamp this down as much as possible.

For example, take a look at https://badssl.com. Many of the red-color links there lead to pages with questionable security. Chrome will display an interstitial, but users can sometimes bypass this interstisial by clicking on [Advanced]->"Proceed (unsafe)". After the user has bypassed the interstitial, Chrome treats the page as secure, but I am proposing to make an exception for payments-related API. Payment Handler and Payment Request should not work with questionable HTTPS states.

What is the best phrasing to use in this pull request to achieve this goal?

PR #292, opened 2018-04-15
Remove link to issue 157

### LINKED ISSUE #157: Revisiting Payment Method specific filter/capability matching

romandev:
The issue 157 was already closed and it is making a respec error.

closes #157

***
<a href="https://pr-preview.s3.amazonaws.com/romandev/payment-handler/pull/292.html" title="Last updated on Apr 15, 2018, 6:39 AM GMT (2a0bbc2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-handler/292/d9ebc35...romandev:2a0bbc2.html" title="Last updated on Apr 15, 2018, 6:39 AM GMT (2a0bbc2)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1052,13 +1052,6 @@
           handler and return the result.
           </li>
         </ol>
-        <div class="issue" data-number="157">
-          This is the current thinking about filtering of payment instruments.
-          As we continue to experiment with implementations, we are soliciting
-          feedback on this approach. Also see the <a href=
-          "https://github.com/w3c/payment-request-info/wiki/PaymentMethodPractice#canmakepaymentevent-implementation">
-          Wiki page</a>.
-        </div>
         <section id="capabilities-example" class="informative">
           <h2>
             How to specify capabilities


rsolomakhin:
The note should stay in, but it does not have to link to issue #157. The reason for keeping the note is to inform the reader that the `CanMakePaymentEvent` is still under discussion.

index.html
```diff
@@ -1052,13 +1052,6 @@ <h2>
           handler and return the result.
           </li>
         </ol>
-        <div class="issue" data-number="157">
```
rsolomakhin, line 1055: Make change to `<div class="note">` instead?

rsolomakhin (changes requested):

index.html
marcoscaceres: Notes should be things we plan to keep in the spec tho.

ianbjacobs:
Closing; subsumed by a different pull request

PR #293, opened 2018-04-15
Add [SameObject] to paymentManager attribute

romandev:
We don't need to create a new object when accessing paymentManager
attribute. So, we just add `[SameObject]`.

Implementation commitment:

 * [x] Chrome (http://crrev.com/c/1013660)

***
<a href="https://pr-preview.s3.amazonaws.com/romandev/payment-handler/pull/293.html" title="Last updated on Apr 16, 2018, 3:39 PM GMT (fd8a710)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-handler/293/d9ebc35...romandev:fd8a710.html" title="Last updated on Apr 16, 2018, 3:39 PM GMT (fd8a710)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -306,7 +306,7 @@
         </p>
         <pre class="idl">
         partial interface ServiceWorkerRegistration {
-          readonly attribute PaymentManager paymentManager;
+          [SameObject] readonly attribute PaymentManager paymentManager;
         };
         </pre>
         <p>


rsolomakhin:
Can you also expose `paymentManager` in Worker context, please? This allows service workers to update instruments in background, as was our intention.

romandev:
@rsolomakhin I removed `[SecureContext]` in this patch. I mis-understood something. There is no reason to add `[SecureContext, Exposed=(Window, Worker)]` to partial interface. The original interface has already the extended attributes. partial interface's extended attribute should be a subset of original interface.

rsolomakhin:
@romandev: I could not access the payment manager in service worker context unless I added `[Exposed=(Window, Worker)]` on the PaymentManager interface. No need to add `Exposed` to the partial interface attribute. Could you do that, please?

```webidl
[
    Exposed=(Window,Worker), // Please add this line.
    RuntimeEnabled=PaymentApp,
    ConstructorCallWith=ExecutionContext
] interface PaymentManager {
    readonly attribute PaymentInstruments instruments;
    attribute DOMString userHint;
};
```

https://cs.chromium.org/chromium/src/third_party/blink/renderer/modules/payments/payment_manager.idl?rcl=ef6fc42bdc561722ddb2c0fbb0eec6d2f1d568ba&l=13

romandev:
@rsolomakhin, We already have it in current spec :)
https://w3c.github.io/payment-handler/#paymentmanager-interface

The `ServiceWorkerRegistration` has already `[SecureContext, Exposed=(Window, Worker)]` attributes.
Therefore, its members are also `[SecureContext, Exposed=(Window, Worker)]` unless we define it.
Am I missing something?

rsolomakhin (approved):
Ah, it's the bug in Chromium. We need to add `[Exposed=(Window,Worker)]` in `payment_manager.idl` to match the spec.

PR #377, opened 2020-08-18
Editorial: Align with Web IDL specification

autokagami:
🤖 This is an automated pull request to align the spec with the latest Web IDL specification. 🤖

The followings are the Web IDL validation messages, which may help understanding this PR:

* ```
  Validation error at line 5 in payment-handler,1:
              Promise<void> enableDelegations(sequence
                      ^
  ```
  
  > Error: `void` is now replaced by `undefined`. Refer to the [relevant GitHub issue](https://github.com/heycam/webidl/issues/60) for more information.
* ```
  Validation error at line 7 in payment-handler,3:
              Promise<void> set(DOMString
                      ^
  ```
  
  > Error: `void` is now replaced by `undefined`. Refer to the [relevant GitHub issue](https://github.com/heycam/webidl/issues/60) for more information.
* ```
  Validation error at line 8 in payment-handler,3:
              Promise<void> clear()
                      ^
  ```
  
  > Error: `void` is now replaced by `undefined`. Refer to the [relevant GitHub issue](https://github.com/heycam/webidl/issues/60) for more information.
* ```
  Validation error at line 7 in payment-handler,7:
              void respondWith(Promise<
              ^
  ```
  
  > Error: `void` is now replaced by `undefined`. Refer to the [relevant GitHub issue](https://github.com/heycam/webidl/issues/60) for more information.
* ```
  Validation error at line 18 in payment-handler,11:
            void respondWith(Promise<
            ^
  ```
  
  > Error: `void` is now replaced by `undefined`. Refer to the [relevant GitHub issue](https://github.com/heycam/webidl/issues/60) for more information.

Currently this autofix might introduce awkward code formatting, and feel free to manually fix it whenever it happens.

Please file an issue at https://github.com/saschanaz/webidl-updater/issues/new if you think this is invalid or should be enhanced.


<!--
    This comment and the below content is programatically generated.
    You may add a comma-separated list of anchors you'd like a
    direct link to below (e.g. #idl-serializers, #idl-sequence):

    Don't remove this comment or modify anything below this line.
    If you don't want a preview generated for this pull request,
    just replace the whole of this comment's content by "no preview"
    and remove what's below.
-->
***
[...]

PR #384, opened 2021-04-01
Replace void with undefined in Web IDL

foolip:
See https://github.com/heycam/webidl/pull/906.


foolip:
@saschanaz FYI

romandev (approved):
lgtm

saschanaz:
(#377 is there)

Issue #401, opened 2022-07-22
Request for use cases: "canmakepayment" event


rsolomakhin:
Hi,

As part of looking into improving the privacy properties of Payment Handlers, I'd like to collect the known use cases for the `"canmakepayment"` event. Here is what I have came up with so far:

1. **Kill switch**: A payment app wants to globally disable the use of Web Payment API. A service worker may use the `“canmakepayment”` event for this purpose, in case if there is a bug that causes the Web Payment API flow to be broken.
1. **Single instance**: A payment app wants to prevent showing two instances of the payment UI at the same time. Chrome rejects any `PaymentRequest.show()` calls that happen while another `PaymentRequest.show()` call is happening, but there is no way for web developers to know that the reject is going to happen ahead of time. A service worker may use the `“canmakepayment”` event to check an in-memory boolean variable that is set to `“true”` when this service worker is already handling an ongoing `"paymentrequest"` event. (This implementation would not catch another origin's Payment Handler UI being displayed, however.)
1. **Instrument on file**: A payment app wants to check whether the user is logged in and has a payment instrument on file.

I'm assuming that the same use cases hold for [Android payment apps](https://web.dev/android-payment-apps-developers-guide/#androidmanifest.xml).

What other use cases are out there? I'd love to hear feedback.

Thank you,
Rouslan

stephenmcgruer:
cc @ianbjacobs - perhaps something for the agenda for either the 18th or TPAC? (Perhaps as a broader discussion of PaymentHandlers in a more-privacy-preserving world, and what the future should look like?)

ianbjacobs:
Hi @stephenmcgruer and @rsolomakhin, yes, let's start by putting this on the 18 August agenda. It would be great for you to re-introduce the topic then.

PR #404, opened 2022-11-15
[Spec] Remove user-identifiable information from canMakePayment

stephenmcgruer:
The “canmakepayment” service worker event lets the merchant know whether the user has a card on file in an installed payment app. It silently passes the merchant's origin and arbitrary data to a service worker from payment app origin. This cross-origin communication happens on PaymentRequest construction in JavaScript, does not require a user gesture, and does not show any user interface. As such, it is a potential source of silent user tracking in a post-3p cookies world.

See #401 for discussions around use-cases for the canmakepayment event.

 * [x] web platform tests (link) - N/A (tests are sadly currently in a poor enough state that the entire test suite needs rewritten)
 * [x] MDN Docs added - N/A? (I cannot locate PaymentHandler docs on MDN...)

Implementation commitment:

 * [x] Safari - N/A, does not ship PaymentHandler
 * [x] Chrome ([link to issue](https://groups.google.com/u/1/a/chromium.org/g/blink-dev/c/AM2bwKxXacQ))
 * [x] FirefoxN/A, does not ship PaymentRequest or PaymentHandler
 * [ ] Edge (public signal)

***
<a href="https://pr-preview.s3.amazonaws.com/stephenmcgruer/payment-handler/pull/404.html" title="Last updated on Dec 14, 2022, 12:37 PM UTC (5750402)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/payment-handler/404/d61d41c...stephenmcgruer:5750402.html" title="Last updated on Dec 14, 2022, 12:37 PM UTC (5750402)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1397,7 +1397,7 @@
                 Algorithm</a>.
               </dd>
               <dt>
-                <a data-lt="CanMakePaymentEvent.modifiers">modifiers</a>
+                <a data-lt="PaymentRequestEvent.modifiers">modifiers</a>
               </dt>
               <dd>
                 The result of executing the <a>Modifiers Population
--- a/index.html
+++ b/index.html
@@ -816,18 +816,10 @@
         <pre class="idl">
           [Exposed=ServiceWorker]
           interface CanMakePaymentEvent : ExtendableEvent {
-            constructor(DOMString type, optional CanMakePaymentEventInit eventInitDict = {});
-            readonly attribute USVString topOrigin;
-            readonly attribute USVString paymentRequestOrigin;
-            readonly attribute FrozenArray&lt;PaymentMethodData&gt; methodData;
+            constructor(DOMString type);
             undefined respondWith(Promise&lt;boolean&gt; canMakePaymentResponse);
           };
         </pre>
-        <p>
-          The <dfn>topOrigin</dfn>, <dfn>paymentRequestOrigin</dfn>,
-          <dfn>methodData</dfn>, and <dfn>modifiers</dfn> members share their
-          definitions with those defined for {{PaymentRequestEvent}}.
-        </p>
         <section>
           <h2>
             <dfn>respondWith()</dfn> method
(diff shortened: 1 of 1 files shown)


stephenmcgruer:
(I do not appear to be able to set reviewers in this repository, so cc @rsolomakhin @ianbjacobs for review).

rsolomakhin (approved):

romandev (approved):
lgtm

ianbjacobs (approved):

stephenmcgruer:
This was discussed in the WPWG today ([minutes](https://www.w3.org/2022/12/08-wpwg-minutes.html#t03)). We heard no objections and therefore I believe it is clear to land - once I update the WPT tests. (Any day now... 🤣 )

stephenmcgruer:
@ianbjacobs @rsolomakhin - ok, after a few days of looking into the tests, I'm sorry to say I think I need to ask that we merge this without updating the WPT tests. The reason is that the test suite is in such a sorry state that it is currently *so far* from passing even without this change today, which makes testing this fairly impossible without re-writing the entire test suite to fix all the other problems.

A short list of problem I found *so far*:

- Tests treat `supportedMethods` as a list still, not a string
- Tests assert that `additionalDisplayItems` and `total` are not in modifiers for "canmakepaymentevent", which they are in Chrome's implementation
- Tests assert that `modifier.data` exists for "canmakepaymentevent", and it doesn't seem to for Chrome
- Tests presume that "canmakepaymentevent" affects canMakePayment(), not hasEnrolledInstrument()
- Tests presume that if "canmakepaymentevent" returns false, then show() will throw a NotSupportedError
- Tests try to reject the "canmakepaymentevent" `respondWith` with Error objects that have actual messages, which is not supported afaik (maybe they silently become false?)

Some of these may be actual bugs in Chrome, whilst some reflect how both the PaymentHandler spec and test suite have diverged from changes in both PaymentRequest spec and in the reality of the Chrome implementation. In all worlds, the test suite (and spec) desperately needs some TLC to at least reflect the shipping reality. But for now, I'd like to just merge this change and address that larger concern later.

(I will open a follow-on issue about the test suite)

rsolomakhin:
> merge this without updating the WPT tests

This sounds reasonable given your explanation. If the other reviewers are OK, too, then I can merge this.

PR #427, opened 2026-02-17
[Editorial] Rename 'payment handler' to 'web-based payment handler' in the spec text

stephenmcgruer:
closes #???

 * [ ] web platform tests (link)
 * [ ] MDN Docs added (link)

Implementation commitment:

 * [ ] Safari (link to issue)
 * [ ] Chrome (link to issue)
 * [ ] Firefox (link to issue)
 * [ ] Edge (public signal)

### DIFF
--- a/index.html
+++ b/index.html
@@ -72,7 +72,7 @@
         };
     </script>
   </head>
-  <body data-cite="service-workers payment-method-id">
+  <body data-cite="service-workers payment-method-id payment-request">
     <section id="abstract">
       <p>
         This specification defines capabilities that enable Web applications to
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -72,7 +72,7 @@
         };
     </script>
   </head>
-  <body data-cite="service-workers payment-method-id">
+  <body data-cite="service-workers payment-method-id payment-request">
```
stephenmcgruer, line 75: I'm not sure if this is the right way to do it, but this stopped respec from complaining that the `[=payment handler=]` reference didn't exist.

stephenmcgruer:
I renamed almost all of the `payment handler` mentions to `web-based payment handler`, whether or not they were a link or just text. It is possible that some of them either (1) should be link-ified and/or (2) should be named back to just 'payment handler' (e.g., if they're talking about the generic concept). Feel free to suggest places that should be changed :)

ianbjacobs (approved):

ianbjacobs:
(I deleted my comment; I left questions on a different pull request.)

index.html
romandev: This looks correct to me. In the long term, migrating to `xref` would be preferable (per [respec guidance](https://respec.org/docs/#data-cite)), but that's a larger change unrelated to this patch. I think this is the right approach for now.

romandev (approved):

index.html
stephenmcgruer: Ack, thanks!

